PR #9, opened 2020-09-09
add docker agent and push images to docker

### LINKED ISSUE #8: Set up CI/CD pipeline for test workbench
- build docker image for master branch and push it to GCR with tag _LATEST_
- retrigger Kubernetes deployment

Out of Scope for now:
- different stages (dev vs, prod)
- using SHA in image name to better control which image gets deployed

pihme:
closes #8

This implements a part of #8, namely the pushing of Docker images

### DIFF
--- a/.ci/podSpecs/distribution.yml
+++ b/.ci/podSpecs/distribution.yml
@@ -39,4 +39,20 @@ spec:
         - name: shared-data
           mountPath: /home/shared
           mountPropagation: Bidirectional
-
+    - name: docker
+      image: docker:18.09.4-dind
+      args: ["--storage-driver=overlay2"]
+      securityContext:
+        privileged: true
+      tty: true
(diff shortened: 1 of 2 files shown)


pihme:
@menski Would appreciate your review. This works, but it prints out the secret to the build log, which is not ideal. Not sure if there is a different way to do this,

Also for the second part - retriggering Kubernetes deployment - I don't know if I need a different Jenkins agent or where to start.

menski (approved):
Looks good to me, I added a fix to prevent the secret to be logged to the console. Not sure why the normal Jenkins masking of secrets did not work. 

Regarding the deployment we can maybe get inspiration from the chaos tests https://github.com/zeebe-io/zeebe/blob/develop/.ci/pipelines/chaos_test_zeebe.groovy#L47-L68

PR #17, opened 2020-09-21
Add process to create cluster and wait for it to be ready

### LINKED ISSUE #15: Add process to create cluster in cloud and wait for it becoming healthy
Currently a situation can arise where clusters are being created but not cleaned up in Camunda cloud.

This can happen in the following circumstances:
* Cluster is not ready in time and task to create the cluster is repeated
* Testbench pod is restarted between creating the cluster and the cluster becoming healthy

Desired change:
* Create a new process that creates a cluster in Camunda cloud and the waits for it to become healthy
* Split up the existing "create cluster in camunda cloud" task into creation and checking the state
* Use process instead of task

pihme:
This PR adds a new process that creates a cluster in the cloud and waits till it is ready. This process replaces what was before a single step. The step had the disadvantage that it might be interrupted and then the cluster would not be cleaned up properly.

closes #15


core/src/main/java/io/zeebe/clustertestbench/worker/QueryClusterStateInCamundaCloudWorker.java
```diff
@@ -0,0 +1,75 @@
...
+		this.cloudClient = new CloudAPIClientFactory().createCloudAPIClient(cloudApiUrl,
+				cloudApiAuthenticationServerURL, cloudApiAudience, cloudApiClientId, cloudApiClientSecret);
+	}
+
+	@Override
+	public void handle(JobClient client, ActivatedJob job) throws Exception {
+		final Input input = job.getVariablesAsType(Input.class);
+
+		String clusterStatus = Optional.of(cloudClient.getClusterInfo(input.getClusterId())).map(ClusterInfo::getStatus)
+				.map(ClusterStatus::getReady).orElse("Unknown");
```
saig0, line 33: I'm not sure which of the return value can be `null`. If `getClusterInfo()` can return `null` then you must use `Optional.ofNullable()` instead.

saig0 (approved):
I like it :cake: 

We could improve the workflow (later) to handle cases when the cluster is not created in time. For example, send a (slack) notification if the cluster is not created within one hour.

pihme:
bors retry

PR #18, opened 2020-09-22
add daily test protocol

### LINKED ISSUE #16: Process to run tests every day
Create a process that runs the tests every day. The process shall run indefinitely until it receives a cancel message

pihme:
This PR adds a process that runs tests on a daily basis.

closes #16


workflows/daily-test-protocol.bpmn
```diff
@@ -0,0 +1,127 @@
...
+        <zeebe:ioMapping>
+          <zeebe:input source="= &#34;new chaos&#34;" target="region" />
+          <zeebe:input source="=&#34;Internal Dev&#34;" target="channel" />
+          <zeebe:input source="=[&#34;Development&#34;, &#34;Production - S&#34;, &#34;Production - M&#34;, &#34;Production - L&#34;]" target="clusterPlans" />
+          <zeebe:input source="={&#34;steps&#34;:3,&#34;iterations&#34;:10,&#34;maxTimeForIteration&#34;:&#34;PT10S&#34;,&#34;maxTimeForCompleteTest&#34;:&#34;PT2M&#34;}" target="sequentialTestParams" />
+        </zeebe:ioMapping>
+      </bpmn:extensionElements>
+      <bpmn:incoming>Flow_1ne8ttl</bpmn:incoming>
+      <bpmn:outgoing>Flow_17xbukq</bpmn:outgoing>
+    </bpmn:serviceTask>
```
saig0, line 20: It seems that the tests are triggered by a service task but I can't find a job worker with this type. 

Or, do you want to use call activity?

saig0: In the diagram, it would be nice to see the parameters of the tests. Maybe, we can add these parameters as an annotation to the task.

workflows/daily-test-protocol.bpmn
```diff
@@ -0,0 +1,127 @@
...
+      <bpmn:outgoing>Flow_1xdnvfy</bpmn:outgoing>
+      <bpmn:outgoing>Flow_06x7sd0</bpmn:outgoing>
+    </bpmn:eventBasedGateway>
+    <bpmn:intermediateCatchEvent id="timer-one-day" name="24h">
+      <bpmn:incoming>Flow_1xdnvfy</bpmn:incoming>
+      <bpmn:outgoing>Flow_13ib8z9</bpmn:outgoing>
+      <bpmn:timerEventDefinition id="TimerEventDefinition_0b3jk4o">
+        <bpmn:timeDuration xsi:type="bpmn:tFormalExpression">PT24H</bpmn:timeDuration>
+      </bpmn:timerEventDefinition>
+    </bpmn:intermediateCatchEvent>
```
saig0, line 33: Is it important when the tests are running?

Maybe, we want to run the tests on a given timeslot (e.g. in the night). Currently, the tests are running always when the workflow instance is created.

workflows/daily-test-protocol.bpmn
```diff
@@ -0,0 +1,127 @@
...
+    <bpmn:sequenceFlow id="Flow_1xdnvfy" sourceRef="Gateway_15hiy9b" targetRef="timer-one-day" />
+    <bpmn:intermediateCatchEvent id="stop-daily-tests-message-event" name="Stop Daily Test">
+      <bpmn:incoming>Flow_06x7sd0</bpmn:incoming>
+      <bpmn:outgoing>Flow_1to83qg</bpmn:outgoing>
+      <bpmn:messageEventDefinition id="msg-stop-daily-test-event" messageRef="msg-stop-daily-test" />
+    </bpmn:intermediateCatchEvent>
```
saig0, line 39: Do you also plan to interrupt the current tests? 

What are the benefits compared to a cancel command?

workflows/daily-test-protocol.bpmn
saig0: What happens if something goes wrong in one day? For example, the cluster can not be created.

Do we want to stop or continue as normal on the next day?

saig0 (commented):
I like the idea :candy: 

I have a few questions about the workflow. Let's have a chat f2f.

workflows/daily-test-protocol.bpmn
pihme: fixed

workflows/daily-test-protocol.bpmn
pihme: fixed

PR #41, opened 2020-10-02
40 technical documentation

### LINKED ISSUE #40: add technical documentation

pihme:
closes #40


Zelldon:
Hey @pihme what do you want me to review actually ? Only the technical documentation part or more? Like the README. Are in general other things already reviewed by other people?

pihme:
@Zelldon Just the documents please

docs/technical-documentation.md
```diff
@@ -0,0 +1,125 @@
...
+Solution Strategy
+=================
+
+The _Zeebe cluster testbench_ is basically a bunch of workers that request jobs from the _testbench_ test orchestration cluster.
+
+The test orchestration cluster contains different workflows that define the steps for a test. See [README.md](../README.md) for a documentation of the different processes. 
+
+Once a test is started, this will e.g. trigger the worker to create a cluster. Afterwards, another worker is triggered to run the test. Finally, other workers are triggered to publish the test results, notify engineers (if needed) and destroy the cluster again. 
+
+The workers are stateless. The only state is kept by Zeebe in the _testbench_ test orchestration cluster. Workers exchange information by reading from and writing to variables in the workflow.
```
Zelldon, line 48: Can we shorten this: `_testbench_ test orchestration cluster` I have read this now multiple times. Maybe just testbench or TTO cluster or something else.

docs/technical-documentation.md
```diff
@@ -0,0 +1,125 @@
...
+![System Scope](assets/system-context.svg "System context")
+
+Solution Strategy
+=================
+
+The _Zeebe cluster testbench_ is basically a bunch of workers that request jobs from the _testbench_ test orchestration cluster.
+
+The test orchestration cluster contains different workflows that define the steps for a test. See [README.md](../README.md) for a documentation of the different processes. 
+
+Once a test is started, this will e.g. trigger the worker to create a cluster. Afterwards, another worker is triggered to run the test. Finally, other workers are triggered to publish the test results, notify engineers (if needed) and destroy the cluster again.
```
Zelldon, line 46: Actually the workers are not really triggered since they poll, but idk how accurate we want to be here.

docs/technical-documentation.md
```diff
@@ -0,0 +1,125 @@
...
+1. Register workers 
+
+Fault Tolerance/Recovery
+------------------------
+Fault tolerance and recovery are handled by the _testbench_ test orchestration cluster:
+ * Whenever a task fails, it will be retried. 
+ * If a worker dies, the job will time out and be reactivated.
+ * If the application crashes, all workers die. As soon as the application is back up, workers can poll for job again.
+ 
+ The _testbench_ test orchestration cluster is deployed to a high availability cluster.
```
Zelldon, line 80: which cluster plan are we using? and which version?

docs/technical-documentation.md
```diff
@@ -0,0 +1,125 @@
...
+* No framework (yet). Mostly to keep the dependencies to a minimum and not to commit on any architectural pattern too soon. (Given that the current implementation is mostly workers, a reactive, non-blocking IO framework would be ideal. However, it is questionable whether the load will ever get so high that the benefits of those frameworks materialize. CDI has been missed while implementing the current solution. So developer convenience might be a stronger driver for architectural commitment then technical criteria) 
+* All environment variables that are used by the solution are read out in the bootstrap class. Mostly because it gets intransparant when environment variables are sprinkled throughout the code.
+* Environment variables are currently the only way to configure the application
+* Each worker defines its input and output parameters in dedicated classes. this is a little more verbose than necessary, but it also documents their interface.
+* The communication to the _Cloud API_ uses RESTEasy Client API. This was the option with the least dependencies (in comparison to Spring RestTemplate, MicroProfile RestClient, and other )
```
Zelldon, line 89: Can we make out of that a separate module/repo?

docs/technical-documentation.md
```diff
@@ -0,0 +1,125 @@
...
+* All environment variables that are used by the solution are read out in the bootstrap class. Mostly because it gets intransparant when environment variables are sprinkled throughout the code.
+* Environment variables are currently the only way to configure the application
+* Each worker defines its input and output parameters in dedicated classes. this is a little more verbose than necessary, but it also documents their interface.
+* The communication to the _Cloud API_ uses RESTEasy Client API. This was the option with the least dependencies (in comparison to Spring RestTemplate, MicroProfile RestClient, and other )  
+
+Risks and Technical Debts 
+=========================
+* Currently, there is no orderly shutdown. So far this has not caused any problems. However, it does slow the tests down when the application crashes or is restarted due to redeploys. The delay is due to Zeebe for the jobs to timeout before rescheduling new jobs.
```
Zelldon, line 93: What is the default activation timeout we are using?

docs/technical-documentation.md
```diff
@@ -0,0 +1,125 @@
...
+* All workers shre the same thread pool. If this thread pool dies or is blocked, then nothing will move forward.
```
Zelldon, line 94:
```suggestion
* All workers share the same thread pool. If this thread pool dies or is blocked, then nothing will move forward.
```

docs/technical-documentation.md
```diff
@@ -0,0 +1,125 @@
...
+* The application has no self monitoring. It relies heavily on Zeebe to restart jobs when things go wrong.
```
Zelldon, line 95: Are we logging things?

docs/technical-documentation.md
```diff
@@ -0,0 +1,125 @@
...
+* The process to create a cluster has a potential infinite loop. If the cluster is created, but never gets ready, then the process will not terminate.
+* The worker `MapNamesToUUIDsWorker` writes to the same variables that it uses as input. Once overwritten, it is no longer possible to look at what the input was. This already confused root cause analysis for a bug.
```
Zelldon, line 97: We could change that right ? :)

docs/technical-documentation.md
```diff
@@ -0,0 +1,125 @@
...
+* The _CloudAPI_ might change and become incompatible. Diagnosing these problems is quite tricky with the RESTEasy Client API. It does a great job at giving developers a nice interface; however if a given REST endpoint returns e.g. a HTML page with a helpful human readable error message it is difficult to get a hold of this error message.
```
Zelldon, line 98: I would expect that we then also get a http error status code, no?

docs/technical-documentation.md
```diff
@@ -0,0 +1,125 @@
...
+========
+
+Glossary 
+--------
+
+
+| Term | Definition |
+| --- | --- |
+| Cloud API | API provided by Camunda Cloud to create, query and destroy clusters, and to create, query and delete client accounts for these clusters |
+| _testbench_ test orchestration cluster | The Zeebe cluster in which the tests are orchestrated |
```
Zelldon, line 110: Pleaseee lets shorten this :D

docs/technical-documentation.md
```diff
@@ -0,0 +1,125 @@
...
+| test driver | Set of classes to run a test and determine the outcome of the test |
+| worker | Class that handles Zeebe jobs. Workers are registered for _service tasks_ of a given job type. | 
+
+
+Arc42 template
+--------------
+
+This document takes inspiration from arc42, the Template for documentation of software and system
```
Zelldon, line 118: I like it :+1:

docs/technical-documentation.md
```diff
@@ -0,0 +1,125 @@
...
+
+The test orchestration cluster contains different workflows that define the steps for a test. See [README.md](../README.md) for a documentation of the different processes. 
+
+Once a test is started, this will e.g. trigger the worker to create a cluster. Afterwards, another worker is triggered to run the test. Finally, other workers are triggered to publish the test results, notify engineers (if needed) and destroy the cluster again. 
+
+The workers are stateless. The only state is kept by Zeebe in the _testbench_ test orchestration cluster. Workers exchange information by reading from and writing to variables in the workflow.
+
+Building Block View
+===================
+![System Whitebox](assets/system-whitebox.svg "System Whitebox")
```
Zelldon, line 52: How is the notification actually working? it seems this is also a worker? But as far as I know it also reported when a workflow was not completed in time? Or there was an exception how is this working?

Zelldon (commented):
Hey @pihme 

cool stuff thanks for doing that :rocket: I added some comments and questions below :)

docs/technical-documentation.md
pihme: not that accurate

docs/technical-documentation.md
pihme: Production - S and Zeebe 0.24.3 (but subject to upgrades)

docs/technical-documentation.md
pihme: It is already a separate module. Code quality is not good enough to make it a prominent repo, IMHO. I will introduce the testbench to the cloud team some day and mention that we have a Java client implementation for their API. If there is interest, we could move it to a dedicated repo

docs/technical-documentation.md
pihme: depends on the worker; ranges from seconds to 30m

docs/technical-documentation.md
pihme: Sure

docs/technical-documentation.md
pihme: Not exactly. I think you get an exception. Not sure whether the exception is "Response could not be be unmarshalled" or more like "regauest returned status XX"

docs/technical-documentation.md
pihme: The test driver detects failures - be it exceptions or maximum times being exceeded.
The test driver collects failures in the test report and stores the test report as variable in the test orchestration workflow. 
The notify worker reads out that variable from the workflow and writes it into the sheet.

docs/technical-documentation.md
Zelldon: Would be nice to have it. I could then use that maybe for the chaos experiments as well. Or we automate the chaos experiments with the test bench :thinking:

docs/technical-documentation.md
Zelldon: Nice thanks :+1:

docs/technical-documentation.md
Zelldon: Do we want to add this to the doc or do you thing it will change to frequently?

docs/technical-documentation.md
Zelldon: Do you think we need this high timeout?

docs/technical-documentation.md
pihme: I don't want to add this to the doc. The documentation should not describe the deployment, IMO. 

Also, a customer could use the testbench and follow the setup guide and do their own tests.

docs/technical-documentation.md
pihme: I re-eneabled Maven Upload on the build. So your snapshots of the master build. If you need a fixed release version, this could be done as well.

docs/technical-documentation.md
pihme: Potentially yes. The high timeout is for the sequential test. The length depends on the test parameters. Currently the tests run only 2 minutes or so, but if we every want to have longer running tests, we need this or even higher timeouts. I also thought about registering a worker twice - one with a low timeout and once with a high timeout and then the test process could have logic which worker to use. For now, the 30 minutes is a bit of an arbitrary middle of the road number so you could have tests that run a little longer.

Zelldon (approved):
Thanks Peter

PR #68, opened 2020-10-23
Clarify the messages that must be published

saig0:
* show the name of the message that should be published to the workflow instead of the id in the BPMN
* the id of the message in the BPMN is not used outside of the BPMN


pihme:
many thanks @saig0!

pihme (approved):

PR #85, opened 2020-11-02
feat(core): externalize Slack channel to env var

### LINKED ISSUE #84: Externalize slack channel to environment variable
Currently the slack channel to publish to is stored in the header of the service task in the workflow description.

The downside of this is that one cannot easily change the channel to publish to.

In particular, it would be desirable to have a dev stage where publishing is directed to a different channel.

This issue is to remove the slack channel from the header and instead set it as an environment variable.

pihme:
closes #84


Zelldon (approved):
bors r+ :+1:

PR #99, opened 2020-11-06
fix(workflows): add missing variable to workflow

### LINKED ISSUE #98: chaos tests cannot be run when called from qa protocol workflow
* the chaos tests cannot be run when called from the qa protocol workflow. This is because a required variable is not set.

+ Mention that the Zeebe image must have a certain syntax to be accepted by the backend.

pihme:
closes #98


pihme:
@Zelldon just some smaller changes. I created a PR in case you already made up your mind whether you want to use or remove the `chaosTestParams`

Zelldon (approved):

Issue #105, opened 2020-11-09
Chaos worker expects wrong cluster plan name format


Zelldon:
Worker expected format `Production-S`, but it is actually `Production - S`. This means experiments are not found.

```
bash-5.0# cat zeebe-chaos/chaos-experiments/camunda-cloud/output-20201109.log 
From https://github.com/zeebe-io/zeebe-chaos
 * branch            master     -> FETCH_HEAD
Already up to date.
[2020-11-09 08:48:08 ERROR] Path "development/*/experiment.json" does not exist.
From https://github.com/zeebe-io/zeebe-chaos
 * branch            master     -> FETCH_HEAD
Already up to date.
[2020-11-09 08:49:54 ERROR] Path "production - s/*/experiment.json" does not exist.
From https://github.com/zeebe-io/zeebe-chaos
 * branch            master     -> FETCH_HEAD
Already up to date.
[2020-11-09 08:50:16 ERROR] Path "production - m/*/experiment.json" does not exist.
From https://github.com/zeebe-io/zeebe-chaos
 * branch            master     -> FETCH_HEAD
Already up to date.

```

PR #106, opened 2020-11-09
chore(chaos): extract io mapping to util script

Zelldon:
Input extraction is moved into separate util script, which is then also used later for output creating.
 Util script is imported by handler. Tests are written against util functions

closes https://github.com/zeebe-io/zeebe-cluster-testbench/issues/105

### DIFF
--- a/core/chaos-workers/handleJobTest.sh
+++ b/core/chaos-workers/handleJobTest.sh
@@ -9,11 +9,17 @@
 
 
 variables="{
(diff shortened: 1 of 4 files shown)


core/chaos-workers/ioHandlerUtil.sh
```diff
@@ -0,0 +1,39 @@
...
+
+
+extractZeebeAddress() {
+  variables=$1
+  echo "$variables" | jq -r '.authenticationDetails.contactPoint'
+}
+
+extractTargetNamespace() {
+  variables=$1
+  echo $(echo "$variables" |  jq -r '.authenticationDetails.audience' | cut -d'.' -f 1)-zeebe
```
pihme, line 38: I would rather take the value from `clusterId`

core/chaos-workers/handleJobTest.sh
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bats
+
+. ioHandlerUtil.sh
+
+@test "read standard in" {
+  result=$(echo "HELLO" | readStandardIn)
+  [ "$result" == "HELLO" ]
```
pihme, line 7: I think it would make more sens to exit on failures. And then to call this test script from the Jenkins pipeline and check for system exit code

pihme (approved):
Some suggestions for improement

core/chaos-workers/handleJobTest.sh
Zelldon: this is done by the lib :)

Zelldon: I will later if I have more time integrate that in our pipeline.

PR #109, opened 2020-11-09
feat(core): improve error message when items cannot be found

### LINKED ISSUE #108: Improve error message when generation could not be found
Currently the error message when a generation could not be found dumps all information about all generations. 

That's too much information. It would be more helpful to just print out the possible names.

pihme:
closes #108

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
 
 	<parent>
@@ -18,8 +20,6 @@
 	<name>Zeebe Cluster Testbench - Core</name>
 
 	<properties>
-		<version.jackson>2.11.2</version.jackson>
-		<version.resteasy>4.5.6.Final</version.resteasy>
 		<version.jaxrs>3.0.12.Final</version.jaxrs>
 		<version.google-api>1.30.11</version.google-api>
 		<version.google-sheets-api>v4-rev20200922-1.30.10</version.google-sheets-api>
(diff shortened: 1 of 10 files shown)


core/pom.xml
```diff
@@ -86,6 +86,16 @@
 			<version>${version.google-http-client}</version>
 		</dependency>
 
+		<dependency>
+			<groupId>io.vavr</groupId>
+			<artifactId>vavr</artifactId>
```
Zelldon, line 91: I have read about this lib before. Did you used it before?

core/src/main/java/io/zeebe/clustertestbench/handler/CreateGenerationInCamundaCloudHandler.java
```diff
@@ -89,34 +87,28 @@ protected static String createGenerationName(String zeebeImage) {
...
 
-    if (optTemplateGenerationInfo.isEmpty()) {
-      throw new RuntimeException("Unable to find generation: " + generationTemplate);
-    }
-    return optTemplateGenerationInfo.get();
+    return eitherTemplateGenerationInfo.getOrElseThrow(msg -> new IllegalArgumentException(msg));
   }
 
-  private Optional<GenerationInfo> findGenerationInfoByName(final String name) {
+  private Either<String, GenerationInfo> findGenerationInfoByName(final String name) {
```
Zelldon, line 95: Could we replace our either in zeebe with vavr? Probably ?

core/src/main/java/io/zeebe/clustertestbench/util/StringLookup.java
```diff
@@ -0,0 +1,103 @@
...
+   * @param candidates the candidates to be searched
+   * @param extractor function that extracts the relevant string from a candidate
+   * @param ignoreCase flag whether the search should respect or ignore case
+   */
+  public StringLookup(
+      String entityToFind,
+      String searchTerm,
+      List<T> candidates,
+      Function<T, String> extractor,
+      boolean ignoreCase) {
```
Zelldon, line 53: You don't need to change it is just a comment:

I'm not a fan of boolean flags, ideally we had something speaking either a enum or classes which indicate the behavior. Like CaseSensitiveLookup and InSensitiveLookup, internally they have a composition of the StringLookup logic and use the different prefixes.

Zelldon (approved):
Thanks :+1:  I was in the beginning not sure what we want to solve :D But I think I get it now :+1:

core/pom.xml
pihme: Not yet. That's why I want to use it here. I find it easier to introduce dependencies and use them intensively in side projects first.

core/src/main/java/io/zeebe/clustertestbench/handler/CreateGenerationInCamundaCloudHandler.java
pihme: We could, but I as far as I remember Nico and Phil evaluated it in the beginning and decided against it.

Issue #107, opened 2020-11-09
Execute bash unit tests from jenkins


Zelldon:
For the chaos worker (written in bash scripts) we have added some unit tests, which should be executed by out jenkins pipeline.

Zelldon:
Cool would also if can run [shellcheck](https://github.com/koalaman/shellcheck) and fail the build if it found something. With the last PR https://github.com/zeebe-io/zeebe-cluster-testbench/pull/111 I cleaned up the scripts

PR #112, opened 2020-11-10
chore(ci): verify chaos scripts

Zelldon:
Run shellcheck for all chaos bash scripts.
 Run bash unit tests in jenkins pipeline.
 Add a separate stage to verify the shell scripts

Dependencies are installed directly from repository, I tried some hours with debians package manager and the different backport channels etc but had no luck / didn't worked.

closes https://github.com/zeebe-io/zeebe-cluster-testbench/issues/107

### DIFF
--- a/.ci/scripts/distribution/prepare.sh
+++ b/.ci/scripts/distribution/prepare.sh
@@ -1,4 +1,15 @@
 #!/bin/sh -eux
 
 apt-get -qq update
-apt-get install --no-install-recommends -qq -y jq libatomic1
+apt-get install --no-install-recommends -qq -y jq libatomic1 bats git
+
+# install shell check
+shellcheckVersion=v0.7.1
+wget "https://github.com/koalaman/shellcheck/releases/download/$shellcheckVersion/shellcheck-$shellcheckVersion.linux.x86_64.tar.xz"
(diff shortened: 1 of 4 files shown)


.ci/scripts/distribution/prepare.sh
```diff
@@ -1,4 +1,15 @@
...
+# install shell check
+shellcheckVersion=v0.7.1
+wget "https://github.com/koalaman/shellcheck/releases/download/$shellcheckVersion/shellcheck-$shellcheckVersion.linux.x86_64.tar.xz"
+tar -xvf "shellcheck-$shellcheckVersion.linux.x86_64.tar.xz"
+cp "shellcheck-$shellcheckVersion"/shellcheck /usr/bin/shellcheck
+
+# install bats
+git clone https://github.com/bats-core/bats-core.git
+cd bats-core
+./install.sh /usr/local
```
pihme, line 15: Do we need a newline and end of file?

pihme (approved):
🎉 Looks great

.ci/scripts/distribution/prepare.sh
Zelldon: hm saw no problem without it :see_no_evil:

Zelldon: I fixed it

Issue #88, opened 2020-11-03
Implement TestResult for Chaos test


Zelldon:
# Description

With https://github.com/zeebe-io/zeebe-cluster-testbench/pull/86 we got a base chaos worker, but it just completes the job with `testResult: PASSED`. It should reflect the real result of the experiments. If a test fails the results should contain at least the name of the failing experiment.

PR #113, opened 2020-11-10
chore(chaos): complete job with fail message

Zelldon:
Build correct testResult json with chaos experiment name as content

closes https://github.com/zeebe-io/zeebe-cluster-testbench/issues/88

### DIFF
--- a/core/chaos-workers/handleJobTest.sh
+++ b/core/chaos-workers/handleJobTest.sh
@@ -102,3 +102,57 @@ failFunction() {
 
   echo "expected"
 }
+
+@test "create failure message without args" {
+  # given
+  expected="$(jq -n '{testResult: "FAILED", failureMessages: [], failureCount: 1, metaData: {}}')"
+
+  # when
+  failureMsg=$(createFailureMessage)
+
+  # then
+  echo "$failureMsg"
+  echo "expected: $expected"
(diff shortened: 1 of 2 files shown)


core/chaos-workers/handleJobTest.sh
```diff
@@ -102,3 +102,57 @@ failFunction() {
...
+  # then
+  echo "$failureMsg"
+  echo "expected: $expected"
+  [ "$failureMsg" == "$expected" ]
+}
+
+
+@test "create failure message with multiple fail messages" {
+  # given
+  expected="$(jq -n '{testResult: "FAILED", failureMessages: ["2", "hallo"], failureCount: 1, metaData: {}}')"
```
pihme, line 148: In this case the optimal result would be `failureCount=2`

pihme (approved):
Looks good to me. Just one comment on the call with two parameters, which you currently don't use anyways. So feel free to ignore it.

core/chaos-workers/handleJobTest.sh
Zelldon: 2 because I have two messages ? Was not sure how this field is used.

pihme: That's more or less the idea yes. But I agree that one could also have one failure and several messages for it.

PR #116, opened 2020-11-10
chore(ci): add optional deployment to dev stage

### LINKED ISSUE #100: Add dev stage to build pipeline
It would be nice if we could test changes before deploying them.

Adapt build pipeline to:
- add a build parameter whether changes should be deployed to dev environment
- build a dev image
- deploy dev image to dev environment

pihme:
closes #100 

This adds a build parameter "DEPLOY_TO_DEV"

When this is set:
* the image will be tagged with `:dev`
* all artifacts will be deployed into the namespace `testbench-dev`
* for all artifacts the deployment descriptor `...-dev.yaml` will be used

When this is not set and we are on the master, it will continue deploy the `:latest` image to `testbench` namespace as usual.

Enabling "DEPLOY_TO_DEV" on master is also possible and will deploy to dev. One use case of this could be to overwrite a bad deployment on dev.

### DIFF
--- a/.ci/scripts/deploy.sh
+++ b/.ci/scripts/deploy.sh
@@ -18,7 +18,7 @@ elif [ "${tag}" = "dev" ]; then
   echo "Deploying :dev to 'dev' stage / 'testbench-dev' namespace"  
   suffix="-dev"
 else 
-  echo "Unknown tag '${tag}'"
+  echo "Unknown tag '${tag}'. Please provide the tag 'dev' or 'latest'"
   exit 1
 fi  
 
--- a/.ci/scripts/deploy.sh
+++ b/.ci/scripts/deploy.sh
@@ -2,11 +2,36 @@
 
 set -ex
 
+if [ -z "$1" ]
+then
+  echo "Please provide the tag 'dev' or 'latest'"
+  exit 1
+fi
+
+
+tag=$1
+
+if [ "${tag}" = "latest" ]; then
+  echo "Deploying :latest to 'int' stage / 'testbench' namespace"  
+  suffix=""
+elif [ "${tag}" = "dev" ]; then
+  echo "Deploying :dev to 'dev' stage / 'testbench-dev' namespace"  
+  suffix="-dev"
+else 
+  echo "Unknown tag '${tag}'"
+  exit 1
+fi  
+
+echo "suffix: ${suffix}"
+
+namespace="testbench${suffix}"
+echo "target namespace: ${namespace}"
+
 gcloud config set core/project zeebe-io
 gcloud config set compute/region europe-west1
 gcloud config set compute/zone europe-west1-b
 
-set +x; echo ${SA_CREDENTIALS} > sa-credentials.json; set -x
+set +x; echo "${SA_CREDENTIALS}" > sa-credentials.json; set -x
 
 gcloud auth activate-service-account jenkins-ci-cd@zeebe-io.iam.gserviceaccount.com --key-file=sa-credentials.json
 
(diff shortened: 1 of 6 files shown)


.ci/scripts/deploy.sh
```diff
@@ -2,11 +2,36 @@
...
+if [ -z "$1" ]
+then
+  echo "Please provide the tag 'dev' or 'latest'"
+  exit 1
+fi
+
+
+tag=$1
+
+if [ "${tag}" = "latest" ]; then
```
Zelldon, line 14: first I was wondering whether this works as a comparison but it seems so.

.ci/scripts/deploy.sh
```diff
@@ -2,11 +2,36 @@
...
+  echo "Deploying :latest to 'int' stage / 'testbench' namespace"  
+  suffix=""
+elif [ "${tag}" = "dev" ]; then
+  echo "Deploying :dev to 'dev' stage / 'testbench-dev' namespace"  
+  suffix="-dev"
+else 
+  echo "Unknown tag '${tag}'"
```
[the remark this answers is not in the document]
Zelldon: Alternatively you could assign per default `suffix=""` and only check whether tag is eq to "dev" and then replace it. Just as an idea.

Jenkinsfile
```diff
@@ -122,25 +132,25 @@ pipeline {
...
                 withVault([vaultSecrets: [
                         [path: 'secret/common/ci-zeebe/zeebe-chaos-service-account', secretValues: [
                                 [vaultKey: 'token'],
                         ]],
                 ]]) {
 
                   dir('core/chaos-workers/') {
-                    sh 'docker build . -t gcr.io/zeebe-io/zeebe-cluster-testbench-chaos:latest --build-arg TOKEN=${token}'
-                    sh 'docker push gcr.io/zeebe-io/zeebe-cluster-testbench-chaos:latest'
+                    sh "docker build . -t gcr.io/zeebe-io/zeebe-cluster-testbench-chaos:${env.TAG} --build-arg TOKEN=${env.token}"
```
Zelldon, line 144: Why we need these env prefix? How does it work? Sorry noob question :D

Jenkinsfile
```diff
@@ -113,7 +114,16 @@ pipeline {
...
 
     stage('Deploy') {
-    	when { branch 'master' }
+    	when { 
+    	   anyOf {
+    	       branch 'master'
+    	       expression { params.DEPLOY_TO_DEV }
+    	   } 
+    	}
+
```
Zelldon, line 123: I think we need a editorcfg for this jenkins file. We could copy the one we have from the zeebe repo.

Zelldon (commented):
Looks good! Just smaller questions/comments.

core/chaos-workers/chaosWorker-dev.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: apps/v1
```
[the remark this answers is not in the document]
pihme: It is what I came up with. The specs differ in 
* the referenced image
* the env var settings

I think you can use something like kustomize to make sure that you only maintain the changes. I didn't have the appetite to try this, though. But some mechanism like this could help to avoid duplicate specs.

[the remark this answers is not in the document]
Zelldon: Yeah true. I agree it should be fine for now thanks :+1:

Jenkinsfile
pihme: Maybe you don't. I am not 100% sure myself.

In this version the string is interpolated by Groovy/Jenkins not by bash. In Jenkins  this is the most common way to refer to environment variables, afaik.

You could probably also do:
`sh 'docker build . -t gcr.io/zeebe-io/zeebe-cluster-testbench-chaos:${TAG} --build-arg TOKEN=${token}'`

(spoiler: the difference is in the `'` vs. `"` around the string)

Then Jenkins would pass this String to bash and bash would do the interpolation.

I like the Jenkins interpolation a little more, because it will give you warnings if something does not exist. I think bash will always resolve it to an empty string, but I might be wrong.

But, to be honest, I have no preference on which way to solve this problem. This was the first way I came up with and it worked. 

And that's why we do a review, so you can look at it and express your preferences :-)

Jenkinsfile
pihme: What is that?

Zelldon: * Our editorcfg https://github.com/zeebe-io/zeebe/blob/develop/.editorconfig
* What is editorcfg https://editorconfig.org/
* Editorcfg eclipse https://marketplace.eclipse.org/content/editorconfig-eclipse

Jenkinsfile
Zelldon: Thanks @pihme, just that I get it right. With `"` jenkins will do the interpolation and here it needs `env.token`? Not sure what my preference would be I was used to it, without prefix. If we go one way we should be consistent this would then be my only concern, but I think you replaced already everything right?

pihme: > Thanks @pihme, just that I get it right. With " jenkins will do the interpolation and here it needs env.token?

Correct

>  but I think you replaced already everything right?

I wouldn't count on it 🙈 I was pretty much in a "Yay it works" mood yesterday. 

So I need to take a step back again and look at things like consistency.

.editorconfig
```diff
@@ -15,5 +15,5 @@ indent_size = 2
 [*.md]
 indent_size = 2
 
-[*.dsl,*.groovy,Jenkinsfile*]
```
Zelldon: Is this something we need to fix in zeebe as well?!

Zelldon (approved):
Awesome thanks @pihme

.editorconfig
pihme: Yes

Zelldon: Could you create an PR or issue for that please

pihme: Done https://github.com/zeebe-io/zeebe/issues/5842

PR #118, opened 2020-11-12
fix(chaos): use correct json template

### LINKED ISSUE #117: NPE in NotifyHandler
It seems that the failing chaos runs causes some problems in the notify handler

```java
java.lang.NullPointerException at io.zeebe.clustertestbench.handler.NotifyEngineersHandler.composeMessage(NotifyEngineersHandler.java:65) at io.zeebe.clustertestbench.handler.NotifyEngineersHandler.handle(NotifyEngineersHandler.java:24) at io.zeebe.client.impl.worker.JobRunnableFactory.executeJob(JobRunnableFactory.java:44) at io.zeebe.client.impl.worker.JobRunnableFactory.lambda$create$0(JobRunnableFactory.java:39) at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source) at java.base/java.util.concurrent.FutureTask.run(Unknown Source) at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source) at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source) at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source) at java.base/java.lang.Thread.run(Unknown Source)
```

Zelldon:
Fixes wrong json template.

**Before**:
```json
{
		"testResult": "FAILED",
		"failureCount": 1,
		"failureMessages": [],
		"metaData": {
		}
}
```
**After**:

```json
{
	"testResult": "FAILED",
	"testReport": {
		"testResult": "FAILED",
		"failureCount": 1,
		"failureMessages": [],
		"metaData": {
		}
	}
}

```

Ideally we could define the template in one place where java and script clients can take it from.

closes #117

### DIFF
--- a/core/chaos-workers/handleJobTest.sh
+++ b/core/chaos-workers/handleJobTest.sh
@@ -79,7 +79,7 @@ failFunction() {
 @test "run experiment with failing runner and values in array - return FAILED" {
   # given
   array=(1 2)
-  expected="$(jq -n '{testResult: "FAILED", failureMessages: ["1 failed"], failureCount: 1, metaData: {}}')"
+  expected="$(jq -n '{testResult: "FAILED", testReport: {testResult: "FAILED", failureMessages: ["1 failed"], failureCount: 1, metaData: {}}}')"
 
   # when
   result=$(runChaosExperiments failFunction "${array[@]}")
@@ -106,7 +106,7 @@ failFunction() {
 
 @test "create failure message without args" {
   # given
-  expected="$(jq -n '{testResult: "FAILED", failureMessages: [], failureCount: 1, metaData: {}}')"
+  expected="$(jq -n '{testResult: "FAILED", testReport: {testResult: "FAILED", failureMessages: [], failureCount: 1, metaData: {}}}')"
 
   # when
   failureMsg=$(createFailureMessage)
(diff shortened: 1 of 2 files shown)


core/chaos-workers/handleJobTest.sh
```diff
@@ -79,7 +79,7 @@ failFunction() {
 @test "run experiment with failing runner and values in array - return FAILED" {
   # given
   array=(1 2)
-  expected="$(jq -n '{testResult: "FAILED", failureMessages: ["1 failed"], failureCount: 1, metaData: {}}')"
+  expected="$(jq -n '{testResult: "FAILED", testReport: {testResult: "FAILED", failureMessages: ["1 failed"], failureCount: 1, metaData: {}}}')"
```
pihme, line 82: This line and other lines:

I am not aware we have/need a top level `testResult` field. To the best of my recollection the after should be:
```
{
	"testReport": {
		"testResult": "FAILED",
		"failureCount": 1,
		"failureMessages": [],
		"metaData": {
		}
	}
}
```

pihme (changes requested):
Please check where we need a top level `testResult` field.

core/chaos-workers/handleJobTest.sh
Zelldon: Currently we need because otherwise the xor gateway will not work

Zelldon: Of course we can change that, but then lets do it in a separate issue

pihme: So strange. I was under the impression that one of the output bindings takes out the testResult and puts it top level.

pihme (approved):

Issue #139, opened 2020-11-16
Persist chaos experiment logs


Zelldon:
The zbctl handler has the limitation that the sys.output is used to complete the job. On error the sys.error is used to fail the job. 
This means we need to write logs to a file.

Currently this is done to an output file, but it is not persisted to a pvc. This means if the pod is rescheduled the file is lost.
We need to assign a pvc to the chaos worker pod and store the log there.

Issue #87, opened 2020-11-03
Add chaos experiments details to README 


Zelldon:
# Description

 1. Describe in the `Tests` section the Chaos experiments/tests
 2. Add the service task details 
 3. Update the Chaos workflow image
 4. Update the Chaos Test section

PR #140, opened 2020-11-16
Persist chaos log

Zelldon:
As stated in https://github.com/zeebe-io/zeebe-cluster-testbench/issues/139 it is not possible to use the handlers system out to log messages, which is the reason why we need to persist the log to a pvc to be resistant against pod restarts.

Data is stored under: `/home/chaos/data/`

**Example:**

```
bash-5.0# find /home/chaos/data/
/home/chaos/data/
/home/chaos/data/lost+found
/home/chaos/data/output-20201116.log
```

closes https://github.com/zeebe-io/zeebe-cluster-testbench/issues/139

I will documented it as part of https://github.com/zeebe-io/zeebe-cluster-testbench/issues/87

### DIFF
--- a/core/chaos-workers/chaosWorker.yaml
+++ b/core/chaos-workers/chaosWorker.yaml
@@ -1,3 +1,16 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: chaos-data-claim
+  labels:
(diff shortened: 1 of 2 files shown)


pihme (approved):
Looks good from my side. Please have a look at your script checker.

PR #141, opened 2020-11-16
chore(scripts): add script to help a little bit with working with chaos test logs

pihme:
Not proud of it. But took me enough time to find these lines, so I better want to share it.

### DIFF
--- a/scripts/chaos-log-chunker.sh
+++ b/scripts/chaos-log-chunker.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+set -eo pipefail
+
+# this script is a little help when analyzing the logfiles of the chaos tests
+# it looks for the chaos worker pod in the current workspace
+# it then downloads the log and splits the log into individual runs
+# it then checks each run for a failure
+# all failures will appear as *.log-chunk_failed files in the current folder
+
+chaosWorkerPod="$(kubectl get pods -o json -l app=chaos-worker | jq -r '.items[].metadata.name')"
+
+kubectl cp "$chaosWorkerPod":/home/chaos/data/chaostoolkit.log chaostoolkit.log
+
+awk '/#{10,}/{filename=NR".log-chunk"}; {print > filename}' chaostoolkit.log
+
+failedChunks=$(grep -l "failing this experiment" -- *.log-chunk)
+
+for chunk in $failedChunks
+do
+  mv "$chunk" "$chunk-failed"
+done
+
+rm ./*.log-chunk
+
+rm chaostoolkit.log


scripts/chaos-log-chunker.sh
```diff
@@ -0,0 +1,25 @@
...
+# it takes the name of the chaos worker pod as input
+# it then downloads the log and splits the log into individual runs
+# it then checks each run for a failure
+# all failures will appear as *.log-chunk_failed files in the current folder
+
+if [ -z "$1" ]
+then
+  echo "Please provide as first argument the name of the chaos worker pod"
+ exit 1
+fi
```
Zelldon, line 15:
```suggestion
chaosWorkerPod="$(kubectl get pods -o json -l app=chaos-worker | jq -r '.items[].metadata.name')"
```

scripts/chaos-log-chunker.sh
```diff
@@ -0,0 +1,25 @@
...
+
+kubectl cp "$1":/home/chaos/zeebe-chaos/chaos-experiments/camunda-cloud/chaostoolkit.log .
```
Zelldon, line 17:
```suggestion
kubectl cp "$chaosWorkerPod":/home/chaos/zeebe-chaos/chaos-experiments/camunda-cloud/chaostoolkit.log .
```

Zelldon (approved):
Cool stuff @pihme :rocket:

scripts/chaos-log-chunker.sh
```diff
@@ -0,0 +1,21 @@
...
+# it then checks each run for a failure
+# all failures will appear as *.log-chunk_failed files in the current folder
+
+chaosWorkerPod="$(kubectl get pods -o json -l app=chaos-worker | jq -r '.items[].metadata.name')"
+
+kubectl cp "$chaosWorkerPod":/home/chaos/data/chaostoolkit.log chaostoolkit.log
+
+awk '/#{10,}/{filename=NR".log-chunk"}; {print > filename}' chaostoolkit.log
+
+grep -l "failing this experiment" *.log-chunk | awk '{printf "mv %s %s-failed\n", $1, $1}' FS=/ | sh
```
pihme, line 17: @Zelldon I am in a bit of a pickle here. 

This version of the script that does work, but shellcheck doesn't like it. https://ci.zeebe.camunda.cloud/blue/organizations/jenkins/zeebe-io%2Fzeebe-cluster-testbench/detail/script-for-chaos-worker-log/4/pipeline

If I apply the shellcheck suggestion then it doesn't work:
```
 user@BlackBox:/mnt/c/dev/github/zeebe-cluster-testbench/scripts$ ./chaos-log-chunker.sh
tar: removing leading '/' from member names
mv: cannot move '.' to '.-failed': Device or resource busy
mv: cannot move '.' to '.-failed': Device or resource busy
mv: cannot move '.' to '.-failed': Device or resource busy
mv: cannot move '.' to '.-failed': Device or resource busy
```
I guess it is because now awk command is wrong. 

Any ideas how to remedy this?

(I do want to keep shellcheck)

Zelldon: Sorry I missed that I will take a look tomorrow

Zelldon: So I tried with adding `--` before `*.log-chunk` this works, but it will btw not work if there is no error in the log then it just dies silently and will not clean up the chunks.

Maybe it would be easier to just do something like this:


```
failedChunks=$(grep -l "failing this experiment" -- *.log-chunk)

for chunk in $failedChunks
do
  mv "$chunk" "$chunk-failed"
done
```

Zelldon, line 17:
```suggestion
failedChunks=$(grep -l "failing this experiment" -- *.log-chunk)

for chunk in $failedChunks
do
  mv "$chunk" "$chunk-failed"
done
```

Zelldon (changes requested):

scripts/chaos-log-chunker.sh
pihme: Thanks

pihme:
All implemented as per your suggestion. Plus rebased and log compacted. Enjoy!

Zelldon (approved):
awesome :) :+1:

PR #145, opened 2020-11-17
chore(chaos): use runChaos script

Zelldon:
Uses the runChaos script to be more resistant against changes in the zeebe-chaos repo.
Previous it assumed a structure of the repo which might change as we did with zeebe-io/zeebe-chaos#44

### DIFF
--- a/core/chaos-workers/runChaos.sh
+++ b/core/chaos-workers/runChaos.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+# This is a place holder script to make shellcheck happy.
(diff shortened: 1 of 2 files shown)


core/chaos-workers/runChaos.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
pihme, line 1: :-)

pihme (approved):

core/chaos-workers/runChaos.sh
pihme: of course we want to make shellcheck happy ;-)

PR #148, opened 2020-11-17
chore(chaos): configure chaos worker

Zelldon:
Configures the chaos worker:

 * activates one job at a time
 * sets concurrency to 1
 * sets timeout to 1 hour


pihme (approved):
👍

PR #149, opened 2020-11-18
fix(chaos): remove insecure flag

Zelldon:
I mistakenly added yesterday the insecure flag this is the reason why currently the worker doesnt connect :see_no_evil: sorry.


pihme (approved):
I missed it too 🙈

PR #152, opened 2020-11-18
feat(workflows): model test outcomes in process

### LINKED ISSUE #151: Remove default edge from run test workflow
Currently the run test in Camunda work flow checks the test results with two edges:
* testResult == FAILED 
* Default

This is unsafe, because any value that is not FAILED will go into the success branch.

Proposed change:
* remove default branch

pihme:
This removes the default edge after checking the test result.

* With this change, all three outcomes are modeled as separate edges
* `SKIPPED` and `PASSED` are modeled as different edges. In Operate one can see which path was chosen.
* Unknown test results will now produce incidents

closes #151


pihme:
I also experimented with setting the names of edges to make the review easier. Hope it helps.

Zelldon (approved):
Interesting idea :+1:

PR #160, opened 2020-11-24
chore(ci): notify slack when build result changes

pihme:
Changes the Jenkinsfile to notify Slack whenever the build result changes.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -204,16 +204,17 @@ pipeline {
             }
         }
 
-        failure {
+        changed {
             script {
                 if (env.BRANCH_NAME != 'master' || agentDisconnected()) {
                     return
                 }
-
-                slackSend(
-                    channel: '#zeebe-testbench-ci',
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -204,16 +204,17 @@ pipeline {
...
                 }
-
-                slackSend(
-                    channel: '#zeebe-testbench-ci',
-                    message: "Zeebe Cluster Testbench failed on ${env.BRANCH_NAME} for build ${currentBuild.absoluteUrl}"
-            )
+                if (hasBuildResultChanged()) {
+                    slackSend(
+                        channel: '#zeebe-testbench-ci',
+                        message: "Zeebe Cluster Testbench failed on ${env.BRANCH_NAME} for build ${currentBuild.absoluteUrl}"
```
korthout: Don't forget to change the message 🙂

korthout (approved):
Looks good, just one thing, but here's the approval already 👍

Jenkinsfile
pihme: Good point

PR #161, opened 2020-11-24
chore(ci): get secrets from Vault and deploy them automatically

### LINKED ISSUE #128: Deploy secrets as part of the build
Currently the secrets are pre-deployed in the namespaces we deploy to.
The process to deploy the secrets manually is error-prone and inherently unsafe.

There should be an automatic way to:
- take the secrets from vault during a build
- prepare the secrets deployment descriptor with these secrets
- deploy the secrets

pihme:
* removes the secrets template file
* gets secrets from Vault in `Jenkinsfile`
* deploys secrets in `deploy.sh`
* works for both dev and int stage

closes #128


pihme:
I'm so proud of this one. This is so much better than dealing with secrets manually.

Zelldon (approved):
Awesome :tada:

PR #163, opened 2020-11-24
chore(workflows): add root workflow information

### LINKED ISSUE #125: Add variable "root workflow" to daily test run, qa protocol
Currently it is hard to find out which root process led to the execution of a certain test.

This could be remedied by setting a variable "root workflow" (or something like this) in the root workflows.
Current root workflows are:
- Daily Tests
- QA Protocol

pihme:
Adds a new variable `rootWorkflow` to the root workflows. The expectation is that this variable will be inherited to the processes called from the root workflow. This should make it easier in Operate to know which test came from where.

closes #125


Zelldon (approved):
love it - it is sad that we need to do it :see_no_evil:

Issue #122, opened 2020-11-12
Clean up test output variables


pihme:
Currently workers expect the following variables:

- `testResult`
- `testReport` (which contains `testResult`)

This should be changed so that only `testReport`is requried.

Also needs to be changed in the chaos-worker script

PR #164, opened 2020-11-24
chore(workflows): extract testResult from testReport automatically

pihme:
First step towards #122 

(Outdated) This adds an output mapping to the test process to automatically extract the test result from the test report.

Updated: This changes the flow conditions to use the `testResult` inside the `testReport` data, not the top level `testResult` variable

This makes the `testResult` variable obsolete, and then it can be removed from the test processes in later PRs.

### DIFF
--- a/workflows/run-test-in-camunda-cloud.bpmn
+++ b/workflows/run-test-in-camunda-cloud.bpmn
@@ -13,7 +13,7 @@
       <bpmn:outgoing>test_skipped</bpmn:outgoing>
     </bpmn:exclusiveGateway>
     <bpmn:sequenceFlow id="test_failed" name="FAILED" sourceRef="check_test_result" targetRef="notify-engineers">
-      <bpmn:conditionExpression xsi:type="bpmn:tFormalExpression">=testResult = "FAILED"</bpmn:conditionExpression>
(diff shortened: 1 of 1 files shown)


pihme:
@Zelldon Not sure why the change set for the BPMN file is so large. I made a serious attempt to change only one output mapping.

workflows/run-test-in-camunda-cloud.bpmn
```diff
@@ -62,6 +62,9 @@
     <bpmn:callActivity id="Activity_17f0cpn" name="Run Test">
       <bpmn:extensionElements>
         <zeebe:calledElement processId="=testWorkflowId" />
+        <zeebe:ioMapping>
+          <zeebe:output source="=testReport.testResult" target="testResult" />
```
Zelldon, line 66: Not 100% sure but it might cause that we not merge testReport then anymore.

Zelldon: Ok im pretty sure it will not merge the report https://docs.zeebe.io/reference/variables.html?highlight=output#output-mappings

pihme: good catch

pihme:
New approach: This changes the flow conditions to use the `testResult` inside the `testReport` data, not the top level `testResult` variable.

Zelldon (approved):
:+1:

PR #166, opened 2020-11-25
165 incorporate backend changes

### LINKED ISSUE #165: Unable to obtain Zeebe client secrets
There was a change in the backend. This leads to the testbench being unable to obtain client secrets.

pihme:
closes #165 

The changes in the backend were:
* the request to create a client now has a permissions field
* the client secret is only returned as response to the create Zeebe client request
* the old secret returned in the response when asking for information on a client is set to an empty string

Changes in the testbench:
* add permissions to create zeebe client request
* use secret returned from create zeebe client request
* remove obsolete field `secret` from `ZeebeClientConnectionInfo`
* (not related to the bugfix, but necessary to debug what was going on) Add a response filter that throws an exception for _bad request_ responses and copies the request body and response body into the message of the exception

### DIFF
--- a/cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/request/CreateZeebeClientRequest.java
+++ b/cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/request/CreateZeebeClientRequest.java
@@ -1,7 +1,11 @@
 package io.zeebe.clustertestbench.cloud.request;
 
+import java.util.Collections;
+import java.util.List;
+
 public class CreateZeebeClientRequest {
 
+  private static final List<String> PERMISSIONS = Collections.singletonList("zeebe");
   private final String clientName;
 
   public CreateZeebeClientRequest(String clientName) {
@@ -10,6 +14,10 @@ public class CreateZeebeClientRequest {
 
   public String getClientName() {
     return clientName;
+  }
+
+  public List<String> getPermissions() {
+    return PERMISSIONS;
   }
 
   @Override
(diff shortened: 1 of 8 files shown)


cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/request/CreateZeebeClientRequest.java
```diff
@@ -1,7 +1,11 @@
 package io.zeebe.clustertestbench.cloud.request;
 
+import java.util.Collections;
+import java.util.List;
+
 public class CreateZeebeClientRequest {
 
+  private static final List<String> PERMISSIONS = Collections.singletonList("zeebe");
```
Zelldon, line 8: Out of curiosity: What is the difference of List.of("zeebe") and this?

core/src/main/java/io/zeebe/clustertestbench/handler/CreateClusterInCamundaCloudHandler.java
```diff
@@ -57,7 +57,7 @@ public void handle(JobClient client, ActivatedJob job) throws Exception {
 
       client
           .newCompleteCommand(job.getKey())
-          .variables(new Output(connectionInfo, name, clusterId))
+          .variables(new Output(createZeebeClientResponse, connectionInfo, name, clusterId))
```
Zelldon, line 60: You probably want to add this also to the docs? :)

Zelldon (commented):
Looks good :+1: thanks @pihme

Zelldon (approved):

core/src/main/java/io/zeebe/clustertestbench/handler/CreateClusterInCamundaCloudHandler.java
pihme: that's just the constructor. The serialized form has not changed

cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/request/CreateZeebeClientRequest.java
pihme: I cannot think of any difference, tbh. The collections API is older, I think and I am used to it.

Issue #168, opened 2020-11-26
Add branch for chaos tests to skip tests when region!=ultrachaos


pihme:
Currently the chaos tests work only when the cluster is run in the right region/cluster.

This is not a limitation, but by design. This cluster was created so we can run chaos tests without interfering other clusters.

It would be good to add a branch to the processes to check the region and skip the chaos test when the test is run in a different region.

pihme:
There are two places where this could be done:
* inside chaos test process
* inside run all tests in Camunda Cloud

The first has the advantage that we can record that the test was skipped in the test report. It has the disadvantage that we needlessly create a cluster just to skip the test.

The second has the advantage that we don't even enter the chaos test execution and the branch is more visible when looking at the workflow diagram. It has the disadvantage that there is no easy way to record that this test was skipped.

Personally, I prefer the second option. But only by a close margin.

PR #169, opened 2020-11-26
Add conditional branch for non chaos region

Zelldon:
* Adds an exclusive gateway with a condition sequence flow which verifies whether the region is equal to "new chaos"
* Per default the test execution is skipped
* Added an annotation to explain this a bit
closes https://github.com/zeebe-io/zeebe-cluster-testbench/issues/168


pihme (approved):
I like it!

PR #171, opened 2020-11-26
chore(ci): change algo for generation names

pihme:
This changes the algorithm for names for generation that are created by the testbench.

Before:
`temp_[image name]_[UUID]` e.g. `temp_gcr.io/zeebe-io/zeebe:0.26.0-SNAPSHOT-3fbe61a3a766b126e146b098ba09509648ff2df7_94853a44-4229-4e5d-abb2-a19723b3956a`

Downsides:
* the name is massive. It looks terrible in the UI 
image
* the UUID at the end is just so that we can have different generations for the same image (in case different tests are running. Each test is responsible for cleaning up its generation). However, this UUID is not the UUID of the generation (which will be generated after a name has been chosen). So this is confusing and overall a bad idea.
* the image name is also quite long as it contains registry, imagename, version, qualifier and full commit hash. Additionally, it's redundant information because a user could look it up in the UI which image was used. 

Change:
* use a random string instead
* it might be even better if the generation has a nonsensical name. This way no developer will mistakingly chose this generation when creating a cluster manually.

Now:
`temp_[10 alphanumeric digits]` e.g.(`temp_YacMC1F230`)

### DIFF
--- a/core/src/main/java/io/zeebe/clustertestbench/handler/CreateGenerationInCamundaCloudHandler.java
+++ b/core/src/main/java/io/zeebe/clustertestbench/handler/CreateGenerationInCamundaCloudHandler.java
@@ -13,8 +13,8 @@ import io.zeebe.clustertestbench.util.StringLookup;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
-import java.util.UUID;
 import java.util.stream.Collectors;
+import org.apache.commons.lang3.RandomStringUtils;
 
 public class CreateGenerationInCamundaCloudHandler implements JobHandler {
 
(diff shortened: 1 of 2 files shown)


Zelldon:
Small question and a generation which begining of the hash commit doesnt work ? like temp_94853a44

pihme:
It would work if you had a guarantee that the commit hash is in a certain position. For calls to the testbench from Jenkins we know in advance the structure of the image name, because we build the image name.

But to be generic, the image name could be anything. 

Also, if you run two QA runs for the same commit, they would produce the same generation name. And, as mentioned above. I rather have each test have its own generation name.

PR #182, opened 2020-12-03
feat(workflows): add workflow enabling external tools to fetch execution result

### LINKED ISSUE #179: Implement mechanism for Jenkins to query the result of a test run
Find a way that the aggregated test result of a run can be delivered to Jenkins.

Initial idea is to have a service task with a job type that corresponds to a key known to Jenkins (e.g. URL of Jenkins build). Then Jenkins could use zbctl to activate this one job worker and fetch the results.

pihme:
closes #179 

Adds a new process that serves as a bridge between an external tool and Zeebe. The new process has a worker, which needs to be implemented by the external tool. Once implemented, the worker allows the external tool to poll for the result of a workflow execution.


deepthidevaki (approved):
Looks good :+1:

PR #186, opened 2020-12-07
181 remove generation async

### LINKED ISSUE #181: Find way to remove generation in an asynchronous way
Right now the generation is deleted after a test run. However, this assumes that all failing clusters have been analyzed already. This will be changed with #180. After 180 is implemented, a failed cluster might still be pending analysis when the test run completes.

Find a mechanism to remove the generation in an asynchronous way, after the aggregated test result has already been transmitted and after all clusters have been analyzed.

Current idea is to trigger a secondary clean up process that regularly checks how many clusters use the generation. Once this number reaches 0, we delete the generation.

pihme:
closes #181 

Change Set:
* adds worker to trigger another process via start event (as a workaround for the missing _Message Throw Event_)
* adds a worker to check how many clusters use a generation
* adds a new workflow to delete a generation when it is no longer used
* refactors _qa-protocol_ workflow to trigger the workflow to delete a generation at the end of a test
* moves list of workers/service tasks into a dedicated document
* several changes to CI (support integration tests; customize mutation testing config)


pihme:
Sorry, change set has grown more than I had initially anticipated. Reviewing commit by commit is probably the best approach.

.ci/podSpecs/distribution.yml
```diff
@@ -40,18 +42,24 @@ spec:
...
+        - name: DOCKER_TLS_CERTDIR
+          value: ""
       securityContext:
         privileged: true
       tty: true
       resources:
         limits:
-          cpu: 2
-          memory: 2Gi
+          cpu: 8
```
deepthidevaki, line 58: Why do we need to increase the resources?

core/src/test/java/io/zeebe/clustertestbench/handler/CheckGenerationUsageHandlerTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+  CheckGenerationUsageHandler sutHandler;
+
+  @BeforeEach
+  void setUp() {
+    sutHandler = new CheckGenerationUsageHandler(mockCloudAPIClient);
+  }
+
+  @Test
+  void shouldReturnTrueWhenGeenrationIsNotInUse() throws Exception {
```
deepthidevaki, line 72: This must be `shouldReturnFalseWhenGenerationInUse`.

core/src/test/java/io/zeebe/clustertestbench/handler/CheckGenerationUsageHandlerTest.java
```diff
@@ -0,0 +1,180 @@
...
+    final var output = argumentCapture.getValue();
+
+    assertThat(output.isGenerationNotInUse()).isFalse();
+
+    verify(mockCompleteJobCommandStep2).send();
+    verify(mockZeebeFuture).join();
+  }
+
+  @Test
+  void shouldReturnFalseWhenGeenrationIsUsed() throws Exception {
```
deepthidevaki, line 101: This must be `shouldReturnTrueWhenGenerationIsNotUsed`

integration-tests/src/test/java/io/zeebe/clustertestbench/it/handler/TriggerMessageStartEventHandlerIT.java
```diff
@@ -0,0 +1,112 @@
...
+   * This integration test consists of two processes:
+   *
+   * <ul>
+   *   <lI>Main - this process contains a worker to start the secondary process via message start
+   *       event
+   *   <li>Secondary - this process shall be triggered by the main process
+   * </ul>
+   */
+  @Test
+  void shoulTriggerSecondaryProcess() throws InterruptedException {
```
deepthidevaki, line 62:
```suggestion
  void shoulTriggerSecondaryProcess() {
```

integration-tests/src/test/java/io/zeebe/clustertestbench/it/handler/TriggerMessageStartEventHandlerIT.java
```diff
@@ -0,0 +1,112 @@
...
+    // given
+    zeebeClient
+        .newDeployCommand()
+        .addResourceFromClasspath("it/triggermessagestartevent/main.bpmn")
```
deepthidevaki, line 66: Minor: Would it be more maintainable if we use `io.zeebe.model.bpmn` to programatically create test workflows?

pom.xml
```diff
@@ -48,6 +48,20 @@
             <version>${plugin.version.pitest-junit5}</version>
           </dependency>
         </dependencies>
+        <configuration>
+          <mutators>
```
deepthidevaki, line 52: How are these used? I can also wait until your talk tomorrow.

workflows/clean-up-generation.bpmn
```diff
@@ -0,0 +1,120 @@
...
+      </bpmn:extensionElements>
+      <bpmn:incoming>Flow_0g579td</bpmn:incoming>
+      <bpmn:outgoing>Flow_0jopnjj</bpmn:outgoing>
+    </bpmn:serviceTask>
+    <bpmn:sequenceFlow id="Flow_1jflfn5" sourceRef="Event_11qwfg9" targetRef="Gateway_15szw35" />
+    <bpmn:intermediateCatchEvent id="Event_11qwfg9" name="1m">
+      <bpmn:incoming>Flow_0mqld4c</bpmn:incoming>
+      <bpmn:outgoing>Flow_1jflfn5</bpmn:outgoing>
+      <bpmn:timerEventDefinition id="TimerEventDefinition_12g4cld">
+        <bpmn:timeDuration xsi:type="bpmn:tFormalExpression">PT1M</bpmn:timeDuration>
```
deepthidevaki, line 41: Minor: Analysis would take hours to finish. So would it be better to increase the timer?

.ci/podSpecs/distribution.yml
```diff
@@ -26,6 +26,8 @@ spec:
             -XX:+UseContainerSupport
         - name: ZEEBE_CI_SHARED_DATA
           value: /home/shared
+        - name: DOCKER_HOST
+          value: tcp://localhost:2375
```
deepthidevaki, line 30: Why is this required?

deepthidevaki (commented):
Looks good. I have only some minor comments.

.ci/podSpecs/distribution.yml
pihme: This docker container is now running a Zeebe Instance + Zeeqs container as part of the integration test. With less resources this lead to OOMs in the container.

.ci/podSpecs/distribution.yml
pihme: The containers are running side by side. The Maven container runs maven and the integration tests, but doesn't have a Docker daemon available. This connects the Maven container to the Docker container (specified further down in the file).

integration-tests/src/test/java/io/zeebe/clustertestbench/it/handler/TriggerMessageStartEventHandlerIT.java
pihme: Maybe. But I like to be able to look at the workflow. (I cannot yet read the code and imagine the workflow in my head. Plus, having it in modeler also helps to quickly deploy it to a local Zeebe instance and see whether it at least deploys correctly.) I was quite happy working with the modeler.

pom.xml
pihme: Yeah, save your questions for tomorrow. Then at least somebody will ask something :-)

core/src/test/java/io/zeebe/clustertestbench/handler/CheckGenerationUsageHandlerTest.java
pihme: so true

workflows/clean-up-generation.bpmn
pihme: Increased it to 15m

deepthidevaki (approved):
:tada:

PR #188, opened 2020-12-07
feat(core): add businessKey to Google Sheet and Slack message

### LINKED ISSUE #187: Add businessKey to Google Sheet
Currently the Google sheet with test results doesn't tell you which Jenkins job triggered a certain test run. This information is available in the field `businessKey` and should be displayed in the Google sheet as well.

pihme:
closes #187 

* adds the _businessKey_ field to Google Sheet of test reports and Slack message
* the business key for workflows triggered by Jenkins is the URL to the Jenkins build
* the business key for daily tests is empty

### DIFF
--- a/core/src/main/java/io/zeebe/clustertestbench/handler/RecordTestResultHandler.java
+++ b/core/src/main/java/io/zeebe/clustertestbench/handler/RecordTestResultHandler.java
@@ -39,7 +39,7 @@ public class RecordTestResultHandler implements JobHandler {
       "Zeebe Cluster Testbench - Publish Test Results Worker";
   private static final List<String> SCOPES = Collections.singletonList(SheetsScopes.SPREADSHEETS);
 
-  private static final String RANGE = "Sheet1!A1:O";
+  private static final String RANGE = "Sheet1!A1:P";
 
   private final String spreadSheetId;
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/zeebe/clustertestbench/handler/RecordTestResultHandler.java
```diff
@@ -39,7 +39,7 @@
       "Zeebe Cluster Testbench - Publish Test Results Worker";
   private static final List<String> SCOPES = Collections.singletonList(SheetsScopes.SPREADSHEETS);
 
-  private static final String RANGE = "Sheet1!A1:O";
+  private static final String RANGE = "Sheet1!A1:P";
```
deepthidevaki, line 42: I don't understand this. Otherwise this PR looks good. How are you testing this, btw?

pihme: This is the range in the Google Sheet. As we add a column to the sheet we need to extend the range.

This is not tested, to be honest.

deepthidevaki (approved):
:+1:

pihme:
@deepthidevaki Can I take this as an approval?

PR #189, opened 2020-12-08
feat(workflows): move analysis of cluster into dedicated workflow

### LINKED ISSUE #180: Find way to do cluster analysis asynchronously
Right now a test run can only complete and aggregate test results when all failed clusters have been analyzed. This is not ideal for Jenkins-triggered QA runs, where we want to have the result delivered to Jenkins - before a cluster failure has been analyzed.

Find a way to do the analysis in an asynchronous way so that the test can complete, but the cluster is still waiting to be analyzed.

The tricky bit here is to defer the destruction of the cluster till after analysis.

pihme:
closes #180 

Changeset:

* adds new process _Analyse Cluster_
* refactors existing process _Run Test in Camunda Cloud_ to trigger the new process on failure.

Depends on #181


README.md
```diff
@@ -286,6 +287,23 @@ The expectation is that the external tool will register a worker for the job typ
...
 
+##### Analyse Cluster
+
+This process is usually called after a failure occurred. It notifies an engineer and keeps the cluster alive until analysis has completed
+
+![analyse-cluster](docs/assets/analyse-cluster.png)
+
+**Workflow ID:** `analyse-cluster`
+
+| Outputs       | Description                                  | Type         |
```
deepthidevaki, line 298: Outputs or inputs?

deepthidevaki (approved):
LGTM :+1:

PR #199, opened 2021-01-05
chore(core): remove redundant field testResult

### LINKED ISSUE #122: Clean up test output variables

pihme:
closes #122 

Changes:
* Removes the top level field `testResult` from the output. This field is redundant as it is also part of the `testReport`


Zelldon (approved):
:+1: 

Did we changed the xor gw already?

pihme:
> Did we changed the xor gw already?

Yes, months ago

PR #200, opened 2021-01-05
78 occams razor q4 2020

### LINKED ISSUE #78: Occams Razor Q4 2020
This issue is to collect items that shall be cleaned up at the end of the quarter:

* Workers with Cloud API Clients: build client outside and pass it into the class (will also improve testability) ✔️ 
* #70 ✔️ 
* #122 ✔️ 
* See which env vars point to the same thing (e.g. server URLs) ✔️ 
*  update technical documentation (additional workers) ✔️

pihme:
closes #78 

Changes: update to technical documentation and operator guide


docs/operator-handbook.md
```diff
@@ -85,17 +87,21 @@ Setup and deployment:
...
-| ---- | ----- | ------- | 
+| File           | Field              | Content                         |
+| -------------- | ------------------ | ------------------------------- |
 | testbench.yaml | ZCTB_SLACK_CHANNEL | The slack channel to publish to |
-| secrets.yaml | slackToken | The token for your app |
+| ENV_VARS       | SLACK_TOKEN        | The token for your app          |
+
+#### Chaos Experiments
+
+tbd (GCLOUD TOKEN)
```
pihme, line 97: @Zelldon 
Could you add some documentation here on what the steps are to set up the chaos worker? The most important thing I can think of is to obtain a token so that the chaos experiements can talk to the cluster on a k8s level.

If you don't want to have it here, but rather in `chaos-workers/README.md` that's also ok. In that case we could link to the readme from here.

Zelldon: now ? or with a new issue?

docs/operator-handbook.md
```diff
@@ -1,36 +1,38 @@
...
-* Slack channel to recieve notifications  
+
+- Kubernetes cluster
+- Zeebe cluster for test orchestration
+- Camunda Cloud Organizaton account
+- Camunda Cloud API credentials to create test clusters on demand
+- Google Sheet file to receive the test results
+- Servicce account, credentials and permissions to write to that file
+- Slack app and token to send notifications to slack channel
+- Slack channel to recieve notifications
```
Zelldon, line 26: why was it changed from * to - ? Any specific reason for that ? :sweat_smile:

docs/technical-documentation.md
```diff
@@ -51,10 +51,14 @@ The workers are stateless. The only state is kept by Zeebe in the _testbench_ te
 ## Modules
 
 - `core` - contains the launcher and the workers that are orchestrated by the test orchestration cluster
+- `core/chaos-workers` - contains the worker to run chaos experiemnts (implemented in Bash/Python)
```
Zelldon, line 54:
```suggestion
- `core/chaos-workers` - contains the worker to run chaos experiemnts (implemented in Bash)
```

Zelldon (approved):

docs/operator-handbook.md
pihme: As you prefer (my preference would be to do it in this PR if you get to it within the next week)

docs/operator-handbook.md
pihme: I am using VSCode to edit the markdown files. I found it the most pleasant to work with. It is especially nice when you have tables, because it will automatically insert white space so that the columns align nicely in the raw version of the file.
````
| Inputs        | Description                                  | Type         |
| ------------- | -------------------------------------------- | ------------ |
| `generation`  | name of the generation for the cluster       | `String`     |
| `clusterPlan` | name of the cluster plan for the cluster     | `String`     |
| `clusterId`   | ID of the cluster in which Zeebe is tested   | `String`     |
| `clusterName` | Name of the cluster in which Zeebe is tested | `String`     |
| `operateURL`  | URL to Operate web interface                 | `String`     |
| `testReport`  | test report                                  | `TestReport` |
````

One quirk/feature of VSCode is that it applies an opinionated standard to all markdown files. It automatically converts bullets to this. I am not sure why it prefers these over the other, but I like the idea of having a uniform way in all documents.

When working at the documentation this came in really handy. E.g. we had some documents with headings like

```
Heading
=====
```

which then got converted on save to

```
# Heading
```

Zelldon: Sounds good :+1: would be cool if we could have something like that for other ide's is there a editorcfg or something which we can export?

pihme: Not that I am aware of. 
There is https://commonmark.org/ for an attampt at a spec and https://johnmacfarlane.net/babelmark2/ to demonstrate why a spec is needed

Zelldon:
@pihme I added some stuff I hope this helps. Let me know what you think :)

PR #204, opened 2021-01-07
fx(core): replace null values with n/a to prevent misalignment

### LINKED ISSUE #203: When no business key is present, the layout in test report sheet is off
For daily tests the business key is empty. In the test report this leads to the following columns moving up one column to the left and the layout is broken.

This seems to be a general issue with empty cells, and it seems it changed with one of the recent updates to the Sheet API.

Solution:
- have a helper function that translates null values into "n/a" and pass all values through that helper function

pihme:
closes #203 

Changes:
- null values in the test report will now be replaced with "n/a" to preserve column layout

### DIFF
--- a/core/src/main/java/io/zeebe/clustertestbench/handler/RecordTestResultHandler.java
+++ b/core/src/main/java/io/zeebe/clustertestbench/handler/RecordTestResultHandler.java
@@ -30,6 +30,8 @@ import java.util.Locale;
 
 public class RecordTestResultHandler implements JobHandler {
 
+  private static final String NA = "n/a";
+
   private static final DateTimeFormatter INSTANT_FORMATTER =
       DateTimeFormatter.ofLocalizedDateTime(FormatStyle.SHORT)
           .withLocale(Locale.US)
@@ -82,7 +84,7 @@ public class RecordTestResultHandler implements JobHandler {
     client.newCompleteCommand(job.getKey()).send();
   }
 
-  private static List<Object> buildRowDataForSheet(final Input input) {
+  protected static List<Object> buildRowDataForSheet(final Input input) {
     final List<Object> result = new ArrayList<>();
 
     result.add(input.getRegion());
@@ -110,11 +112,13 @@ public class RecordTestResultHandler implements JobHandler {
     result.add(convertMillisToString(testReport.getEndTime()));
     result.add(new ZeebeObjectMapper().toJson(testReport.getMetaData()));
 
+    result.replaceAll(item -> item == null ? NA : item);
+
     return result;
   }
 
   private static String returnValueIfApplicable(final TestReport testReport, final Object value) {
-    return testReport.getTestResult() == TestResult.PASSED ? "n/a" : value.toString();
+    return testReport.getTestResult() == TestResult.PASSED ? NA : value.toString();
   }
 
   private static String convertMillisToString(final long millis) {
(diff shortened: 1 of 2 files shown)


pihme:
@npepinpe Didn't you say you want a little more exposure to the testbench code? What about a little review as a first baby step?

core/src/main/java/io/zeebe/clustertestbench/handler/RecordTestResultHandler.java
```diff
@@ -82,20 +84,20 @@ public void handle(final JobClient client, final ActivatedJob job) throws Except
     client.newCompleteCommand(job.getKey()).send();
   }
 
-  private static List<Object> buildRowDataForSheet(final Input input) {
+  protected static List<Object> buildRowDataForSheet(final Input input) {
```
npepinpe, line 87: I would propose instead to add all the fields, and before sending it, iterate through them and replace null with NA. I think we're less likely to have someone forget to replace null with NA if we do this in the future. We could also extract that to a method, i.e.

```java
protected static List<Object> buildRowDataForSheet(final Input input) {
  final List<Object> result = new ArrayList<>();
  result.add(input.getRegion());
  // ... add more fields

  // null values are omitted from the result sheet, so insert a placeholder in order to preserve the layout
  final List<Object> sanitizedResult = result.stream().map(o -> o == null ? NA : o).collect(Collectors.toList());
  // ...
}
```

core/src/test/java/io/zeebe/clustertestbench/handler/RecordTestResultHandlerTest.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.function.BiConsumer;
+import java.util.stream.Stream;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+class RecordTestResultHandlerTest {
+
+  @ParameterizedTest
+  @MethodSource("getColumsAndSetters")
+  void testShouldWriteValueIntocorrectColumn(
```
npepinpe, line 19:
```suggestion
  void testShouldWriteValueIntoCorrectColumn(
```

core/src/test/java/io/zeebe/clustertestbench/handler/RecordTestResultHandlerTest.java
```diff
@@ -0,0 +1,83 @@
...
+import io.vavr.Tuple3;
+import io.zeebe.clustertestbench.handler.RecordTestResultHandler.Input;
+import io.zeebe.clustertestbench.testdriver.api.TestReport.TestResult;
+import io.zeebe.clustertestbench.testdriver.impl.TestReportDTO;
+import java.util.function.BiConsumer;
+import java.util.stream.Stream;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+class RecordTestResultHandlerTest {
```
npepinpe, line 15: We could add a test that fails if someone adds a field to `Input` (which ends up in the results) but forgets to replace null with NA. All fields in `Input` are by default null, no? Then we create a default Input, pass it, then look if any columns still have null (and fail if so).

npepinpe (changes requested):
Looks good generally, I'm very curious to know why this is the behaviour but I couldn't find anything in the docs (other than their suggestion to add empty strings instead of null if you want to preserve them)

npepinpe (approved):
:tada:

PR #231, opened 2021-02-08
fix(workflows): set max time for cluster creation

pihme:
Changes:
* Adds a timeout to the cluster preparation service task
* If timeout is exceeded, the test is marked as failed and a notification is sent to slack to analyze what went wrong


npepinpe (approved):
Process looks good to me, I didn't run it though - have you tested it in dev? Otherwise I can do it as well.

pihme:
I haven't tested it on dev. Don't know how to test this. How do you test that a cluster takes too long to create?
I did deploy it locally, if you are referring to deployability

npepinpe:
Sorry I worded it poorly - it looks simple enough which is why I didn't, but figured maybe you'd want to be safe and run it anyway to see everything is as expected (e.g. the I/O mappings map properly). To test I would normally just deploy a different version of the workflow with a small time out.

But yeah, it's probably simple enough to skip that :+1:

PR #233, opened 2021-02-09
Revert "fix(workflows): set max time for cluster creation"

pihme:
This reverts commit 28ab16f4f612cb6832fbea5dddf979a69fb30385.

The fix isn't good enough:
- the cluster won't be deleted, because the cluster ID is not transported when the timeout is triggered
- the Slack message is not distinct enough from a normal Slack message
- there is no actionable information


npepinpe (approved):
bors r+

PR #249, opened 2021-03-12
Rename workflow to process everywhere

npepinpe:
**Description**

This PR adapts menski's script provided in https://github.com/camunda-cloud/zeebe/pull/6455 to rename every instance of workflow to process.

The only thing I had to change was the dependency on `zeebe-workflow-generator` which I suppose will stay as is for now.

The other failures are because we use the client, which still refers to workflows, but there has been no release with process yet - so maybe we do an alpha release a little earlier to get that.


npepinpe:
Last issue is that we depend on zeeqs for integration testing, but zeeqs still uses the old API I guess :slightly_smiling_face:

Zelldon:
Plus zeebe-chaos will not work (it still uses workflowInstance as well)

korthout:
rebased on develop branch

korthout:
@pihme Care to take a look at this? I've rebased and added some extra stuff.

korthout:
@npepinpe Please also have a look :)

pom.xml
```diff
@@ -428,7 +428,7 @@
     <version.slf4j>1.7.30</version.slf4j>
     <version.vavr>0.10.3</version.vavr>
     <version.workflowgenerator>1.0.2</version.workflowgenerator>
-    <zeebe.version>0.26.0</zeebe.version>
+    <zeebe.version>1.0.0-alpha4</zeebe.version>
```
npepinpe, line 431:
```suggestion
    <zeebe.version>1.0.0-alpha5</zeebe.version>
```

processes/test-notification.bpmn
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
npepinpe, line 1: I guess we can remove this? I think I used it at some point to test notifications. I must've added it by mistake, sorry

npepinpe (commented):
Looks good to me in general. I take it we disabled the IT tests...maybe depending on zeeqs is the best, as Zeebe will increasingly depend on Testbench in the future, but there are no similar plans for zeeqs (there are some vague plans to have a query API one day though, so that would replace it - I imagine we can wait until more comes out of this)

npepinpe:
As I'm the author, I'm not allowed to approve this PR :smile:

pihme (approved):

processes/test-notification.bpmn
korthout: Removed it from your commit

korthout:
bors merge

Issue #261, opened 2021-03-22
Configure dependabot to also include the stable 0.26 branch


pihme:
Currently dependabot is only looking at the develop branch. It should run for each stable branch:

See a config example: https://github.com/zeebe-io/zeebe-kafka-exporter/blob/master/.github/dependabot.yml

PR #265, opened 2021-03-31
github(dependabot): Configure Dependabot to run also for stable branch

pihme:
Changes dependabot conifguration:
* will now run daily instead of weekly
* will also run for `stable/0.26`
Fixes: #261


korthout (approved):
LGTM 🎁

PR #284, opened 2021-04-09
Improve ChaosToolkit process

Zelldon:
* improve chaosToolkit process
  *  Fix a problem with the failuresMessages, it was expected to be a list instead of a string.
  * Introduce a sub process, since output mappings didn't work for call activities, which makes problems on successful run. The testResult was not updated.

chaosToolkit

* Enables logging of the chaos workers:
  *  Previous the output of the running process was not printed, now it is at least printed to stdout. We can improve here.
* increase memory limits 
  * We migrated to a java application which normally needs more memory.


pihme (approved):
:tada:

Zelldon:
@pihme I sneaked some more changes in I hope this is ok. Mostly add timeout for not waiting for ever on the process which executes the script and increase worker threads.

PR #295, opened 2021-04-15
Use 0.26.1 tag in Chaos Worker

Zelldon:
In order to change the zeebe-chaos experiments, to make them 1.x compatible, we need to use in the chaos worker the old experiments (tagged with 0.26.1). In the Testbench 1.x we can then use the newer experiments which are migrated and on the older Testbench the 0.26.1 version.


korthout (approved):
LGTM 🐵

PR #317, opened 2021-05-06
Switch to 1.0.0-rc1 cluster

korthout:
The switch is made by switching the client-id and by changing the values
in vault.


korthout:
bors merge

PR #318, opened 2021-05-06
fix(testdriver): make sequential process executable

korthout:
When replacing the external dependency to zeebe-workflow-generator, it
seems the wrong method was called to create the sequential test process.


korthout:
bors merge

korthout:
bors retry

PR #320, opened 2021-05-10
feat(processes): analyze cluster as task

### LINKED ISSUE #205: Use tasklist feature for analyze cluster step
Currently we have a process to analyze a cluster. This process uses a message to indicate when analysis has completed. This was a work around since we didn't have human tasks at that time.

Now that the task list is available, we could use a "user task" instead.

korthout:
Since user task is not yet available in the Zeebe modeller, let's add a
service task with type humanTask first. Tasklist can pick this up as a
task to work on.

The previously existing message intermediately catch event has become a
boundary event on the task, so stay backwards compatible.

Resolves #205


pihme (approved):
Please also update the screenshots in the documentation.

korthout:
bors merge

PR #331, opened 2021-05-12
Map chaos test results correctly out

Zelldon:
Previous the test results and especially the failure messages haven't
been propagated correctly for two reasons.

 1. there was no output mapping on the call activities
 2. error events terminate the call activities and can't transfer
    payload

In order to overcome this we need to change the process model such that
it a) maps the payload outside the call activity and b) not terminate
the call activity

Normal run looks like this:

view-running

And with failure it looks like this:

chaosexperiment-fail

If we take a look in the right corner we see the testReport variable which contains all failure messages:

chaostoolkit-fail


pihme (approved):
Love it!

PR #339, opened 2021-05-14
Add on-comment trigger to backport action workflow

korthout:
Adds a workflow configuration to trigger the backport action
when a user writes a `/backport` comment on a pull request.
The bot itself restricts that only merged pull requests can be
backported.

This allows to backport pull requests that have already been merged.
Just add the backport labels to the merged PR and write `/backport`.

> WARNING! The bot will try to backport to the branches of all backport labels.
> Previously created backport pull requests are not taken into account.
> New pull requests will be created for each label.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,14 +1,18 @@
-name: Backport labeled PRs after merge
+name: Backport labeled merged pull requests
 on:
   pull_request:
     types: [ closed ]
+  issue_comment:
+    types: [ created ]
 jobs:
   build:
     name: Create backport PRs
+    if: ${{ github.event_name == 'pull_request' || (github.event_name == 'issue_comment' && github.event.issue.pull_request && contains(github.event.comment.body, '/backport')) }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
         with:
+          # required to find all branches
           fetch-depth: 0
       - name: Create backport PRs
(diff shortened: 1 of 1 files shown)


pihme (changes requested):
Please merge this file with `backport.yml` 

You can combine several triggers in one workflow. See e.g. https://github.com/camunda-community-hub/zeebe-process-generator/blob/8ff462cddb1078c4d842e4176e7ae20b95a02300/.github/workflows/ci.yml#L8

Also would be nice if you could then update:
https://github.com/camunda-community-hub/maven-template/blob/main/.github/workflows/backport.yml (but that is optional)

korthout:
I thought about this before, it was difficult due to the conditional statement. But you're right, I should be able to merge them. I'll have a go at it

pihme:
Yes, it gets a bit clumsy, but you can also check which trigger it was in your if statement. See e.g. https://github.com/camunda-community-hub/zeebe-process-generator/blob/8ff462cddb1078c4d842e4176e7ae20b95a02300/.github/workflows/ci.yml#L49

korthout:
@pihme please have another look. 

It was unclear to me how to deal with the relatively long conditional expression, so I asked about it here: https://stackoverflow.com/questions/67531606/how-to-deal-with-long-conditional-expression-in-github-actions-workflow.

pihme (approved):
Nice!

PR #342, opened 2021-05-14
docs(dev): describe multi version support

### LINKED ISSUE #251: Multi-version support
**Goal**
* have different versions of the testbench running side-by-side. Each version is tied to a Zeebe version (with the expectation that we have one testbench version per major version of Zeebe)

**Steps**
***House Keeping***
* rename master branch to develop
* create branch to capture 0.26 state

***Create new testbench cluster instance for develop***
* maybe in production
* Follow all cluster-specific steps in https://github.com/zeebe-io/zeebe-cluster-testbench/blob/master/docs/operator-handbook.md
* Complete https://github.com/zeebe-io/zeebe-cluster-testbench/pull/249
* Adjust version of referenced zbtcl artifact in chaos workers
* Think about naming conventions for this version and future versions (naming convention could cover: branch, secrets files, namespace in Kubernetes, Docker image names)
* Create secrets in Vault that point to the new cluster (respect naming conventions)
* Create namespace in zeebe-io google cloud for the new workers (respect naming conventions)
* Adjust Jenkinsfile
  * introduce concept of stable branches (very much like we have for Zeebe)
  * retrieve secrets for the right version 
  * Maybe decide on target namespace here and pass it into deployment scripts
* Adjust deployment  scripts (either select right namespace here, or pass in namespace from Jenknsfile)
* Adjust Jenkinsfile in Zeebe
   * Adjust timer trigger: not just for develop, but for develop and stable branches
   * When RUN_QA is set explicitly, we need to figure out which testbench to talk to, based on the Zeebe verssion
   * Pick the right secrets file to talk to the right testbench
* Maybe hard-code image version into daily test protocol BPMN model
* Maybe add version guard to models 
* Add to documentation to explain all this

***Open Issues*** 
* What to do with current dev stage

korthout:
Adds a description about multi-version support to the developer-handbook.

closes #251

### DIFF
--- a/docs/developer-handbook.md
+++ b/docs/developer-handbook.md
@@ -3,12 +3,27 @@
 - [Technical documentation](technical-documentation.md)
 - [Operator handbook](operator-handbook.md)
 
+## Versions
+Testbench orchestration cluster runs on Camunda Cloud and is only compatibile with Zeebe versions that are backwards compatibly supported by the Camunda Cloud version it is running on.
+
+At the time of writing, there are 2 different versions supported by the Testbench: 0.26 and 1.0.
+Zeebe progress on these versions is done in their respective `stable` branches: `stable/0.26` and `stable/1.0`, while new progress is made on `develop`.
+
+Testbench uses the same branching structure to split the version specific code.
+Testbench's `develop` branch is kept compatible with Zeebe's `develop` branch.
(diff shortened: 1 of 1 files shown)


pihme (approved):

korthout:
@pihme Something went wrong with the bot here. Might the folding.

PR #343, opened 2021-05-14
ci(github): fix backport conditions

korthout:
The backport was running unconditionally after #341. It might be that the expression syntax is the cause of this.


korthout:
bors merge

korthout:
test

korthout:
test

PR #355, opened 2021-05-17
fix cloud API url

pihme:
The old URL was from a time when the API was still being developed.
This now has an official URL, this URL is also whitelisted by the firewall, the old URL is not.

This also reformatted one of the markdown files.


pihme:
@korthout FYI

korthout:
@pihme Awesome! Thanks

Zelldon (approved):
:rocket:

korthout:
@pihme Please let me know whether the suggested cherry-pick works for you locally

PR #365, opened 2021-05-20
fix(chaos): replace assembly with shade plugin

Zelldon:
In order to fix our Log4J issue, which was caused by the assembly plugin
and building a jar with dependencies, we need to replace the assembly
plugin with the shade plugin and use a third-party plugin to make sure
that the log4j plugin dat file is correctly merged with all
dependencies.

This makes it possible to have the strackdriver logging and building a
jar with dependencies.

Found it via https://stackoverflow.com/questions/25065580/log4j2-custom-plugins-annotation-processing-with-maven-assembly-plugin and https://maven.apache.org/plugins/maven-shade-plugin/examples/executable-jar.html


pihme (approved):
nice catch :rocket:

PR #366, opened 2021-05-20
Fix docker build

Zelldon:
In order to reference the jar with dependencies this commit gives it a
unique name without artifact or version, which is used in the Dockerfile
as well

Refactored the chaos worker pom a bit:

 * Introduce version properties and migrated to newer version
 * Small clean up of pom (remove unused plugins)


pihme (approved):
Thanks so much for digging so deep to fix this :tada:

Issue #368, opened 2021-05-20
Thread.run blocks the current thread


Zelldon:
I realized while searching through the logs that the standard output is always written first and then the standard error.
Like this:

```
D 2021-05-20T12:08:35.988Z pod/zeebe-0 condition met 
D 2021-05-20T12:08:35.994Z pod/zeebe-1 condition met 
D 2021-05-20T12:09:28.699Z pod/zeebe-2 condition met 
D 2021-05-20T12:09:28.709Z pod/zeebe-3 condition met 
D 2021-05-20T12:09:28.714Z pod/zeebe-4 condition met 
D 2021-05-20T12:09:28.732Z pod/zeebe-5 condition met 
D 2021-05-20T12:09:28.735Z + source utils.sh 
D 2021-05-20T12:09:28.735Z ++ CHAOS_SETUP=cloud 
D 2021-05-20T12:09:28.735Z ++ getNamespace 
D 2021-05-20T12:09:28.735Z ++ '[' -z 5df66ee6-a878-4c98-a8fb-be396a42ab78-zeebe ']' 
D 2021-05-20T12:09:28.735Z ++ namespace=5df66ee6-a878-4c98-a8fb-be396a42ab78-zeebe 
D 2021-05-20T12:09:28.735Z ++ echo 5df66ee6-a878-4c98-a8fb-be396a42ab78-zeebe 
D 2021-05-20T12:09:28.735Z + namespace=5df66ee6-a878-4c98-a8fb-be396a42ab78-zeebe 
D 2021-05-20T12:09:28.735Z + '[' cloud == cloud ']' 
D 2021-05-20T12:09:28.735Z + kubectl wait --for=condition=Ready pod -l app.kubernetes.io/app=zeebe --timeout=900s -n 5df66ee6-a878-4c98-a8fb-be396a42ab78-zeebe 

```
[...]

PR #369, opened 2021-05-20
fix(chaos): print system error together with system out

Zelldon:
Previously the output streams have been printed after another, which can
cause confusion and makes it not really visible what is happening.

Furthermore the worker thread was blocked, until the process was done.
This cause problems on handling jobs, not terminating the process after
some time and we are not seeing what is going on.

For more information read https://github.com/zeebe-io/zeebe-cluster-testbench/issues/368

closes https://github.com/zeebe-io/zeebe-cluster-testbench/issues/368

Sorry with bothering you all the time @pihme

### DIFF
--- a/core/chaos-workers/chaos-worker/src/main/kotlin/io/zeebe/chaos/chaosMain.kt
+++ b/core/chaos-workers/chaos-worker/src/main/kotlin/io/zeebe/chaos/chaosMain.kt
@@ -17,6 +17,7 @@ import java.nio.file.Files
 import java.time.Duration
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
+import kotlin.concurrent.thread
 
 private const val ENV_TESTBENCH_ADDRESS = "TESTBENCH_ADDRESS"
 private const val ENV_TESTBENCH_CLIENT_ID = "TESTBENCH_CLIENT_ID"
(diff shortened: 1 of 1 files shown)


pihme (changes requested):
Hmm, not sure I understood everything. Here is my take on it:

- I don't think logging asynchronously alone is doing what you want it to do. 
- I think you also need to include the `process.waitFor(timeoutInSeconds, TimeUnit.SECONDS)` because this one is also blocking the thread
```
  public boolean waitFor(long timeout, TimeUnit unit) throws InterruptedException {
    long startTime = System.nanoTime();
    long rem = unit.toNanos(timeout);

    while(true) {
      try {
        this.exitValue();
        return true;
      } catch (IllegalThreadStateException var9) {
        if (rem > 0L) {
          Thread.sleep(Math.min(TimeUnit.NANOSECONDS.toMillis(rem) + 1L, 100L));
        }

        rem = unit.toNanos(timeout) - (System.nanoTime() - startTime);
        if (rem <= 0L) {
          return false;
        }
      }
    }
  }
```

- And I guess once you do that you will get the same `IllegalThreadStateException` that the `waitFor` is using - if you call `process.exitValue()` too soon -  so the rest of the method needs to go into the async block as well.
- And now that I think about it, we might want to fail the job or throw an error in the else case, so that Zeebe knows what happened, and doesn't need to wait for the timeout

And then, we have a nice example for an asynchronous worker.

pihme:
Oh and since we are in Kotlin, we might also look at https://kotlinlang.org/docs/coroutines-guide.html

Zelldon:
Yeah I dont wanted to open that box, but I hope soon 😁


The waitFor is ok. I want to block there, there I have a timeout. But on the consuming the output stream I had no timeout which means the process will run forever and we forever consume its output. Forever because the scripts retry on errors. 

I want to consume the output async and want to wait for either exit or timeout. What ever comes first will be then handled. On timeout I destroy the process.

pihme (approved):
lgtm :tada:

PR #379, opened 2021-05-27
Use worker testutils

pihme:
This changes the tests for workers to use the stubs of the https://github.com/camunda-community-hub/zeebe-worker-java-testutils instead of Mocks

### DIFF
--- a/core/src/test/java/io/zeebe/clustertestbench/handler/AggregateTestResultHandlerTest.java
+++ b/core/src/test/java/io/zeebe/clustertestbench/handler/AggregateTestResultHandlerTest.java
@@ -1,32 +1,25 @@
 package io.zeebe.clustertestbench.handler;
 
+import static java.util.Map.entry;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyNoMoreInteractions;
-import static org.mockito.Mockito.when;
+import static org.camunda.community.zeebe.testutils.ZeebeWorkerAssertions.assertThat;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
-import io.camunda.zeebe.client.api.ZeebeFuture;
-import io.camunda.zeebe.client.api.command.CompleteJobCommandStep1;
-import io.camunda.zeebe.client.api.response.ActivatedJob;
-import io.camunda.zeebe.client.api.worker.JobClient;
 import io.vavr.collection.Stream;
 import io.zeebe.clustertestbench.testdriver.api.TestReport.TestResult;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import org.camunda.community.zeebe.testutils.stubs.ActivatedJobStub;
+import org.camunda.community.zeebe.testutils.stubs.JobClientStub;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
-import org.mockito.Mock;
-import org.mockito.Mockito;
-import org.mockito.junit.jupiter.MockitoExtension;
 
-@ExtendWith(MockitoExtension.class)
 public class AggregateTestResultHandlerTest {
 
   static Stream<Arguments> provideValidValues() {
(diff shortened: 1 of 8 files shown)


Zelldon:
@pihme sorry for the long delay. I try to review it today :crossed_fingers:

pihme:
@Zelldon No need to hurry. Review it at your leisure.

core/src/test/java/io/zeebe/clustertestbench/handler/AggregateTestResultHandlerTest.java
```diff
@@ -69,109 +62,71 @@
...
-    private void mockJobCompletChain() {
-      when(mockJobClient.newCompleteCommand(Mockito.anyLong()))
-          .thenReturn(mockCompleteJobCommandStep1);
-      when(mockCompleteJobCommandStep1.variables(Mockito.anyMap()))
-          .thenReturn(mockCompleteJobCommandStep2);
-      when(mockCompleteJobCommandStep2.send()).thenReturn(mockZeebeFuture);
+      assertThat(activatedJobStub)
+          .completed()
+          .extractingOutput()
+          .containsExactly(entry(AggregateTestResultHandler.KEY_AGGREGATED_RESULT, "SKIPPED"));
```
Zelldon, line 129: Love it!

core/src/test/java/io/zeebe/clustertestbench/handler/CheckGenerationUsageHandlerTest.java
```diff
@@ -50,131 +48,74 @@
...
 
-  @SuppressWarnings("unchecked")
-  private void mockJobCompletChain() {
-    when(mockActivatedJob.getKey()).thenReturn(TEST_JOB_KEY);
-    when(mockJobClient.newCompleteCommand(Mockito.anyLong()))
-        .thenReturn(mockCompleteJobCommandStep1);
-    when(mockCompleteJobCommandStep1.variables((Object) Mockito.any()))
-        .thenReturn(mockCompleteJobCommandStep2);
-    when(mockCompleteJobCommandStep2.send()).thenReturn(mockZeebeFuture);
+    assertThat(activatedJobStub).isStillActivated();
```
Zelldon, line 119: Isn't normally a fail command send when an exception is thrown in an job handler?

core/src/test/java/io/zeebe/clustertestbench/handler/DeleteGenerationInCamundaCloudHandlerTest.java
```diff
@@ -64,53 +52,28 @@ void shouldCallApiToDeleteGeneration() throws Exception {
...
-    sutDeleteGenerationhandler.handle(mockJobClient, mockActivatedJob);
+    sutDeleteGenerationHandler.handle(spyJobClientStub, activatedJobStub);
 
     // then
-    final var inOrder = inOrder(mockInternalApiClient, mockJobClient);
+    final var inOrder = inOrder(mockInternalApiClient, spyJobClientStub);
 
     inOrder.verify(mockInternalApiClient).deleteGeneration(Mockito.any());
-    inOrder.verify(mockJobClient).newCompleteCommand(Mockito.anyLong());
+    inOrder.verify(spyJobClientStub).newCompleteCommand(Mockito.anyLong());
```
Zelldon, line 72: why was this mocking in this class not completely removed?

pom.xml
```diff
@@ -1,422 +1,424 @@
...
+
+  <scm>
+    <connection>scm:git:git@github.com:zeebe-io/zeebe-cluster-testbench.git</connection>
+    <developerConnection>scm:git:git@github.com:zeebe-io/zeebe-cluster-testbench.git
+    </developerConnection>
+    <tag>HEAD</tag>
+    <url>https://github.com/zeebe-io/zeebe-cluster-testbench</url>
+  </scm>
+
+  <version>2.0.0-SNAPSHOT</version>
```
Zelldon, line 423: Do you know why this pom is changed like this? Didn't we have a specific format defined for the pom? It is weird to see the version as last tag.

Zelldon (commented):
I like it, looks good :+1: I had just some comments/questions, please have a look.

core/src/test/java/io/zeebe/clustertestbench/handler/CheckGenerationUsageHandlerTest.java
pihme: Yes, but this is not done by the JobHandler. It is done in `JobRunnableFactory.executeJob(...)` 

So as far as unit testing the job handler goes, the job handler leaves the job as it is.

pom.xml
pihme: I think I don't have the right IntelliJ settings. I recently activated "reorganize file on save" because I was annoyed by the Maven errors that the declaration order was incorrect. One side effect of that was that Idea also rearranged POM files. And from what I can tell it sorts everything alphabetically. 

if you have better settings, I am all for it. Otherwise, I need to disable this again.

core/src/test/java/io/zeebe/clustertestbench/handler/DeleteGenerationInCamundaCloudHandlerTest.java
pihme: In this case I am asserting that methods were called in a certain order. This can only be done with mocks,

pom.xml
Zelldon: I tried with https://plugins.jetbrains.com/plugin/12003-pom-format and this works quite good

core/src/test/java/io/zeebe/clustertestbench/handler/DeleteGenerationInCamundaCloudHandlerTest.java
Zelldon: And you think this is necessary? At the end it is just about whether the generation was deleted or not right? But the order of the calls doesn't matter.

Zelldon (approved):
Thanks Peter :+1:

PR #388, opened 2021-06-08
[chaos] Use expressive multi version process name

korthout:
In operate, it is strange to see v1 or v2 as a process name. To make
analysis easier, let's change the name of the multi-version process to
something more expressive


korthout:
@Zelldon note, I found it additionally hard to track this down, since `zeebe-chaos` still contains a shell version of this same experiment in a folder for the same cluster type (L). Should we do something about this duplication?

Zelldon:
Yes I discussed with @pihme already this that we want to move it at some point to the chaos repo. The scripts are still useful during development since you still can run the chaostoolkit so I will keep them for a while. 

You can detect this based on the output whether it is a script or not.

korthout:
closed in favor of #389

PR #389, opened 2021-06-08
Improve multi version deploy

korthout:
It could happen that deployment of the initial version of the
multiVersion process would fail. The handler awaits deployment of the
successive versions and in the end expects that 10 different versions
were deployed, but this could be lower if one of the model_v1
deployments failed.

By reworking the multi version deployment to a collection pipeline, it becomes much more clear 
what happens and when things go wrong. This also allows us to respond to failure cases.
The job can be failed (keeping retries the same) when multi version deployment fails.

This failure should not affect the standard timer boundary event that times out if the task is not 
completed on time (in actionRunner.bpmn).

### DIFF
--- a/core/chaos-workers/chaos-worker/src/main/kotlin/io/zeebe/chaos/DeployMultipleVersionsHandler.kt
+++ b/core/chaos-workers/chaos-worker/src/main/kotlin/io/zeebe/chaos/DeployMultipleVersionsHandler.kt
@@ -8,8 +8,7 @@ import io.camunda.zeebe.client.api.worker.JobHandler
 import io.camunda.zeebe.model.bpmn.Bpmn
 import org.awaitility.kotlin.await
 
-class DeployMultipleVersionsHandler(val createClient: (ActivatedJob) -> ZeebeClient = ::createClientForClusterUnderTest) :
-    JobHandler {
+class DeployMultipleVersionsHandler : JobHandler {
 
     private val PROCESS_ID = "multiVersion"
     private val RESOURCE_NAME = PROCESS_ID +".bpmn"
@@ -20,27 +19,27 @@ class DeployMultipleVersionsHandler(val createClient: (ActivatedJob) -> ZeebeCli
         const val JOB_TYPE = "deploy-different-versions.sh"
     }
 
-    override fun handle(client: JobClient, job: ActivatedJob) {
+    override fun handle(testbench: JobClient, job: ActivatedJob) {
         setMDCForJob(job)
         LOG.info("Handle job $JOB_TYPE")
 
-        createClient(job).use {
-            LOG.info("Connected to ${it.configuration.gatewayAddress}, start deploying multiple versions...")
+        createClientForClusterUnderTest(job).use { clusterUnderTest ->
+            LOG.info("Connected to ${clusterUnderTest.configuration.gatewayAddress}, start deploying multiple versions...")
 
             val lastVersion = IntRange(1, 10)
-                    .map{i -> waitForModelDeployment(it, i)}
+                    .map{i -> waitForModelDeployment(clusterUnderTest, i)}
                     .map{e -> e?.processes?.get(0)?.version ?: -1 }
                     .last()
 
             if (lastVersion < 10) {
                 LOG.warn("Deployed 10 different versions of process $PROCESS_ID, last version: $lastVersion. Fail $JOB_TYPE")
-                client.newFailCommand(job.key)
+                testbench.newFailCommand(job.key)
                         .retries(job.retries)
                         .errorMessage("Expected to deploy 10 different versions of process $PROCESS_ID, but only deployed $lastVersion")
                         .send()
             } else {
                 LOG.info("Deployed 10 different versions of process $PROCESS_ID, last version: $lastVersion. Complete $JOB_TYPE")
-                client.newCompleteCommand(job.key).send()
+                testbench.newCompleteCommand(job.key).send()
             }
         }
     }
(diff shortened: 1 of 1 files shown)


core/chaos-workers/chaos-worker/src/main/kotlin/io/zeebe/chaos/DeployMultipleVersionsHandler.kt
```diff
@@ -31,7 +33,7 @@ class DeployMultipleVersionsHandler(val createClient: (ActivatedJob) -> ZeebeCli
 
             var lastVersion = -1
             for (i in 1..5) {
-                it.deployModel(MODEL_V1, "modelV1.bpmn")
+                await.until { -> it.deployModel(MODEL_V1, "modelV1.bpmn") }
```
Zelldon: Good catch! Maybe we can move the await.until inside? It seems to be always used like that.

core/chaos-workers/chaos-worker/src/main/kotlin/io/zeebe/chaos/DeployMultipleVersionsHandler.kt
```diff
@@ -31,37 +27,31 @@ class DeployMultipleVersionsHandler(val createClient: (ActivatedJob) -> ZeebeCli
...
 
-            var lastVersion = -1
-            for (i in 1..5) {
-                await.until { -> it.deployModel(MODEL_V1, "modelV1.bpmn") }
-                lastVersion = waitForModelDeployment(it, MODEL_V2, "modelV2.bpmn")
-            }
+            val lastVersion = IntRange(1, 10)
+                    .map{i -> waitForModelDeployment(it, i)}
+                    .map{e -> e?.processes?.get(0)?.version ?: -1 }
+                    .last()
```
Zelldon: alternative we could use this:

```kotlin
        createClient(job).use { zeebeClient ->
            LOG.info("Connected to ${zeebeClient.configuration.gatewayAddress}, start deploying multiple versions...")

            val lastVersion = (1..10)
                .map { waitForModelDeployment(zeebeClient, it) }
                .map { it?.processes?.get(0)?.version ?: -1 }
                .last()
```

core/chaos-workers/chaos-worker/src/main/kotlin/io/zeebe/chaos/DeployMultipleVersionsHandler.kt
```diff
@@ -32,10 +32,17 @@ class DeployMultipleVersionsHandler(val createClient: (ActivatedJob) -> ZeebeCli
                     .map{e -> e?.processes?.get(0)?.version ?: -1 }
                     .last()
 
-            LOG.info("Deployed 10 different versions of process $PROCESS_ID, last version: $lastVersion. Complete $JOB_TYPE")
+            if (lastVersion < 10) {
```
Zelldon, line 35: How could this then happen?

core/chaos-workers/chaos-worker/src/main/kotlin/io/zeebe/chaos/DeployMultipleVersionsHandler.kt
```diff
@@ -32,10 +32,17 @@ class DeployMultipleVersionsHandler(val createClient: (ActivatedJob) -> ZeebeCli
...
+                LOG.warn("Deployed 10 different versions of process $PROCESS_ID, last version: $lastVersion. Fail $JOB_TYPE")
```
Zelldon, line 36: This message is then wrong

Zelldon (commented):
Hey @korthout 

thanks so much to dig into it and that you fixed this. I really appreciate it!

I had some comments/questions, would be nice if we can go over it.

core/chaos-workers/chaos-worker/src/main/kotlin/io/zeebe/chaos/DeployMultipleVersionsHandler.kt
korthout: Because of the broker's response. Perhaps something went wrong and it did not create 10 different versions, then it would be `lastVersion < 10 => true`

core/chaos-workers/chaos-worker/src/main/kotlin/io/zeebe/chaos/DeployMultipleVersionsHandler.kt
korthout: Do you still think this is necessary? Since the deploy is now completely defined in side the DeployMultipleVersionsHandler, and that method is no longer used

core/chaos-workers/chaos-worker/src/main/kotlin/io/zeebe/chaos/DeployMultipleVersionsHandler.kt
Zelldon: yeah ok so it need to fail several time until the awaitility stops to retry

core/chaos-workers/chaos-worker/src/main/kotlin/io/zeebe/chaos/DeployMultipleVersionsHandler.kt
Zelldon: yeah you're right. This is the drawback of reviewing commit by commit

Zelldon (approved):
Thanks :+1: don't forget to backport it :)

korthout:
bors merge

PR #390, opened 2021-06-08
Switch analyse cluster to user task

korthout:
It was using a service task with special job type, but that does not appear to work with the current version of Tasklist.

This switches the service task to a user task.


pihme (approved):
Please also update the screenshot of the process in the docs

korthout:
bors merge

Issue #380, opened 2021-05-28
Operate link missing in Slack message


pihme:
The Operate link is no longer visible in the Slack message:

![image](https://user-images.githubusercontent.com/14032870/119942092-bb642e00-bf91-11eb-9fcc-9c648e9d774d.png)

korthout:
I found the culprit: https://github.com/camunda-cloud/console/pull/2066

PR #391, opened 2021-06-10
Log received cluster info

korthout:
It is not always clear what is returned by the cloudAPI. Let's print out
the results at debug level to make it easier to debug when things have
changed.

Specifically, this might help to find out why the operate url is null.

relates to #380

### DIFF
--- a/cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/filter/EntityLoggingFilter.java
+++ b/cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/filter/EntityLoggingFilter.java
@@ -15,6 +15,7 @@ import javax.ws.rs.client.ClientResponseContext;
 import javax.ws.rs.client.ClientResponseFilter;
 import javax.ws.rs.ext.WriterInterceptor;
 import javax.ws.rs.ext.WriterInterceptorContext;
+import org.apache.commons.io.IOUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -75,21 +76,8 @@ public final class EntityLoggingFilter
       final InputStream stream, final ClientRequestContext requestContext) throws IOException {
     final var method = requestContext.getMethod();
     final var uri = requestContext.getUri();
-    final var entity = readEntityFromStream(stream);
+    final var entity = IOUtils.toString(stream, StandardCharsets.UTF_8);
     log(String.format("%s %s => %s%n", method, uri, entity));
-  }
-
-  private String readEntityFromStream(final InputStream stream) throws IOException {
-    final var sb = new StringBuilder();
-    var totalRead = 0;
-    var numRead = 0;
-    var entity = new byte[MAX_ENTITY_SIZE + 1];
-    while ((numRead = stream.read(entity, totalRead, MAX_ENTITY_SIZE)) > 0) {
-      sb.append(new String(entity, 0, Math.min(numRead, MAX_ENTITY_SIZE), StandardCharsets.UTF_8));
-      totalRead += numRead;
-      entity = new byte[MAX_ENTITY_SIZE + 1];
-    }
-    return sb.toString();
   }
 
   private void log(String entry) {
--- a/cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/filter/EntityLoggingFilter.java
+++ b/cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/filter/EntityLoggingFilter.java
@@ -27,7 +27,7 @@ public final class EntityLoggingFilter
 
   private static final Logger LOGGER = LoggerFactory.getLogger(EntityLoggingFilter.class);
   private static final String ENTITY_STREAM_PROPERTY = "EntityLoggingFilter.entityStream";
-  private static final int MAX_ENTITY_SIZE = 1024 * 1024; // original was 1024 * 8 but was too small
+  private static final int MAX_ENTITY_SIZE = 1024 * 8;
 
   @Override
   public void filter(ClientRequestContext requestContext) {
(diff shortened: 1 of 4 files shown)


pihme (changes requested):
I like the idea of logging the requests to be able to diagnose what is going on.

The proposed change is ok, but probably too late in the request processing sequence. This way we will only print out those bits which were successfully deserialized and mapped.

I would propose to use a request filter instead. We already have one for bad requests `BadRequestResponseFilter`. 

The only tricky thing is to mark and reset the input stream, as described here: https://stackoverflow.com/questions/36676017/jax-rs-2-print-json-request

For an all bells and whistles implementation look here: https://github.com/kleben/jersey/blob/master/core-common/src/main/java/org/glassfish/jersey/filter/LoggingFilter.java

But we probably need less than that.

korthout:
The request filter works, (I've tested it on `1.x-dev`), but it's a bit hard to read, since it does not also print the requested url. It also seems to cut-off after about 1k chars and it logs all responses, including all retries for `QueryClusterStatus` and such, leading to relatively many logs produced. Lastly, lgtm found some problems with this code.

The goal of these logs was gain insights into the responded data, but we can also just inspect https://github.com/camunda-cloud/console/blob/master/backend/routes/api/CustomerClusterRouter.ts#L285-L315 for changes. Perhaps we can even come to an understanding with the console team, that they let us know about breaking changes. WDYT?

pihme:
>  but it's a bit hard to read, since it does not also print the requested url. 

I think you can get the the URI from request context

> It also seems to cut-off after about 1k chars 

That's strange.

> and it logs all responses, including all retries for `QueryClusterStatus` and such, leading to relatively many logs produced. 

The many logs doesn't bother me too much. I think we should be able to filter these log lines out because it is a distinct logger. 

> The goal of these logs was gain insights into the responded data, but we can also just inspect https://github.com/camunda-cloud/console/blob/master/backend/routes/api/CustomerClusterRouter.ts#L285-L315 for changes. Perhaps we can even come to an understanding with the console team, that they let us know about breaking changes. WDYT?

I don't think them letting us know about breaking changes is a viable option. I would be hard pressed to define which change is a breaking one. So for them to keep track of what we use and how important it is for us - I think that is too much to ask.

What we might discuss, though, is whether we can provide acceptance tests. These could be JS tests, maintained by either party. And they could run them alongside their other tests - and then alert us when any of the tests breaks. 

Still, I am unsure whether we want to invest that effort given that the breaking changes so far were once a quarter and usually not too bad.

I would definitely prefer to keep the logging class inside the code base. if you are worried about the log volume we could have it inactive, but we would be able to reactivate it when we need to diagnose something

pihme (changes requested):

cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/filter/EntityLoggingFilter.java
```diff
@@ -0,0 +1,112 @@
...
+      final var stream = new LoggingStream(requestContext.getEntityStream());
+      requestContext.setEntityStream(stream);
+      requestContext.setProperty(ENTITY_STREAM_PROPERTY, stream);
+    }
+  }
+
+  @Override
+  public void filter(ClientRequestContext requestContext, ClientResponseContext responseContext)
+      throws IOException {
+    if (responseContext.hasEntity()) {
```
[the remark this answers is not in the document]
korthout: I'd rather not, this was created to log responded entities.

korthout:
@pihme Please have a look at my most recent (5) commits and let me know what you think

cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/filter/EntityLoggingFilter.java
```diff
@@ -0,0 +1,124 @@
...
+    logEntityFromStream(stream, requestContext);
+    stream.reset();
+    return stream;
+  }
+
+  private void logEntityFromStream(
+      final InputStream stream, final ClientRequestContext requestContext) throws IOException {
+    final var method = requestContext.getMethod();
+    final var uri = requestContext.getUri();
+    final var entity = readEntityFromStream(stream);
```
pihme, line 78:
```suggestion
    final var entity =  IOUtils.toString(stream, StandardCharsets.UTF_8);

```

from `org.apache.commons.io`

cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/filter/EntityLoggingFilter.java
```diff
@@ -0,0 +1,124 @@
...
+    log(String.format("%s %s => %s%n", method, uri, entity));
+  }
+
+  private String readEntityFromStream(final InputStream stream) throws IOException {
```
pihme, line 82: I am not an expert on IO streams, but this seems to be wrong in so many ways.

`numRead = stream.read(entity, totalRead, MAX_ENTITY_SIZE)` skips the first `totalRead` bytes in the `entity` array and then starts writing from there, possibly exceeding the size of the array

Then `new String(entity, 0, Math.min(numRead, MAX_ENTITY_SIZE)` copies the array starting at index 0, where maybe nobody has written anything, because of the line above. Depending on `numRead` it might not reach to the point where something was written.

And the whole loop always reads `MAX_ENTITY_SIZE` or less, but since its a loop in total the code might read more than `MAX_ENTITY_SIZE`. 

Also the loop condition I think should be `> -1`. As far as I know, a result of 0 may be okay and simply means the stream has no data yet, but more data might come in in the future. (E.g. if the response is a DVD worth of data, there might be situations where the communication is still ongoig while the receiver starts processing the response).

Anyways, I feel this is getting a bit out of hand for the problem we are trying to solve. The whole business about the `MAX_ENTITY_SIZE` is only a safety measure in case the response is indeed a DVD. Because then the buffered stream would blow memory. 

Since we can expect reasonably short responses, I think it is good enough to copy everything that is there, and there is a utility function which does just that. (see other comment)

pihme (changes requested):
Not sure what happened here. Maybe it's just too hot these days :smile:

cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/filter/EntityLoggingFilter.java
korthout: I just worked from what we had available. I don't think agree with all your comments (although you might be right, I'm no IO Streams expert either and perhaps I misread the docs), but I absolutely agree that simpler is better. So let's just switch to commons.io for this one.

korthout:
@pihme did what you suggested, and played around a bit with the debug level logging configuration. Please have another loook

pihme (approved):
Good stuff :tada:

korthout:
bors merge

PR #392, opened 2021-06-11
Remove daily test protocol process

korthout:
It is replaced by Nightly QA Run started by Jenkins via External Tool
Integration process.

The daily test protocol has been emptied and the name is prefixed with
`Legacy` to indicate its removal, since undeploy is not yet available.

closes #324

PR #394, opened 2021-06-11
Backport 392 to stable/0.26

korthout:
Backports #392 to `stable/0.26`

There were conflicts because of workflow->process renaming, so they were easy to resolve.


korthout:
Since this was very straightforward, I don't think it requires a review.

bors merge

PR #395, opened 2021-06-11
Proof of Concept: generate BPMN models from experiment JSON

pihme:
This is not to be reviewed, more to take note of.

This is a proof of concept for the following idea: Based on the `experiment.json` file we could create a BPMN diagram that represents this experiment. Then we could deploy those diagrams and have even more visibility for the experiments.

Example of generated process:
image

This Proof of Concept shows that it can be done. Otherwise, it is not really fleshed out yet.

Just wanted to test this idea, and know what you think of it.


Zelldon:
Cool idea :+1::rocket: I like it.

Just the generated DI makes me worrying a bit :D

Zelldon (commented):
Cool idea :+1::rocket: I like it.

Just the generated DI makes me worrying a bit :D 

I had no deeper look at the implementation, but looked quite simple :+1:

pihme:
What do you mean by "generated DI"?

korthout (commented):
I like the idea as well 👍 Cool stuff

pihme:
Closing this PR, was only for demonstration. Branch will be kept alive in case we want to come back to it

PR #421, opened 2021-06-24
Fix operate URL in slack message (notify engineers)

### LINKED ISSUE #380: Operate link missing in Slack message

korthout:
The operateURL was no longer provided by the cloud api.
A new property (`links`) was added to ClusterInfo, which this pull request introduces and uses to replace the old one with.

The main work in this pull request is refactoring the schemas for the different api responses. By isolating them (with some duplication), it is easier to fix changes to these individual responses.

closes #380 

Reviewer: please look past the number of lines this pull request changes, it should be very easy to review commit-by-commit.


cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/response/ClusterInfo.java
```diff
@@ -95,41 +79,230 @@ public void setStatus(final ClusterStatus status) {
...
-        + metadata
+        + ", links="
+        + links
         + "]";
   }
 
-  /* TODO
-   * - field "spec"
-   **/
+  @JsonIgnoreProperties(ignoreUnknown = true)
```
pihme, line 113: :thought_balloon: One thing we could reconsider in general is the use of this annotation. 

Back in the first version these were introduced to have some leniency when interpreting the responses. The upside of that leniency was that testbench was able to run without errors even though the Operate link was no longer at the place we expected it. The downside is that we miss schema changes, 

I am still in favor of leniency, personally. I feel it is less stressful when things are running smoothly, and one can take the time to fix any discrepancies that arise. However, I wanted to mention it and put it up for reconsideration.

cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/response/ClusterInfo.java
```diff
@@ -95,41 +79,230 @@ public void setStatus(final ClusterStatus status) {
...
+    public String toString() {
+      return "ChannelInfo [uuid=" + uuid + ", name=" + name + "]";
+    }
+  }
+
+  @JsonIgnoreProperties(ignoreUnknown = true)
+  public static class ClusterStatus {
+
+    private String ready;
+    private String zeebeStatus;
```
pihme, line 229: :thought_balloon: Maybe we can use an enum here in future versions.

cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/response/ClusterInfo.java
```diff
@@ -95,41 +79,230 @@ public void setStatus(final ClusterStatus status) {
...
+          + ready
+          + ", zeebeStatus="
+          + zeebeStatus
+          + ", operateStatus="
+          + operateStatus
+          + "]";
+    }
+
+    /*
+     * TODO - other fields - map status fields to enums
```
pihme, line 268: :question: Do we still need the TODO?

pihme (approved):
Thanks for going the extra mile and updating all DTOs.

:x: Please update the version to 3.0.0- SNAPSHOT (This repo adheres to sementic versioning conventions)
(https://github.com/zeebe-io/zeebe-cluster-testbench/blob/develop/docs/release-process.md)

Another question is whether we want to publish a new release. Here I am undecided. 

We published a release once, because Bernd wanted to use the cloud API jar as dependency in one of his projects. Not sure he still uses it, but if he does, then this update would be helpful. Not sure if he needs an official release, or if a snapshot release would suffice.

Another thought is to extract the cloud API module into its own repo. Then it would be clearer how it should be versioned and released. But it also requires more commitment to maintain it. And it will be harder to make quick changes for testbench. However, it would help in using the updated DTOs in both branches.

My personal preference is to update the version number and not publish a release.

Just wanted to mention the broader context in case you have any ideas or personal aspirations. (a.k.a if you say "This was so much fun, I want to maintain this as a top level project" I wouldn't want to stop you :smile: )

korthout:
>This repo adheres to sementic versioning conventions

I was unaware.

>My personal preference is to update the version number and not publish a release.

I agree. Let's just update the version. If a new release is needed we can always do that on demand.

>Another thought is to extract the cloud API module into its own repo.

This is a nice idea, but out of scope for this changeset IMO.

cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/response/ClusterInfo.java
korthout: That makes sense. Thanks for raising the topic. I'm also in favor in leniency at this time. So let's leave it.

cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/response/ClusterInfo.java
korthout: There's a todo for this :)

cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/response/ClusterInfo.java
korthout: Ah, I can narrow the todo 👍

korthout:
@pihme Just to make sure, I did it correctly. I've updated the version using `mvn versions:set`. Was that correct?

korthout:
bors merge

Issue #423, opened 2021-06-30
Remove 0.26 Testbench


Zelldon:
With the newest 1.1.0 release we no longer support 0.26, which means we can remove our Testbench 0.26

Zelldon:
Lets summarize what we need to do here @pihme @korthout I will write down things I think we need to do maybe you have more or different approach.

 * delete stable branch (0.26)
 * delete jenkins job (?)
 * delete testbench 0.26 namespace in zeebe gke
 * delete testbench 0.26 Camunda cloud cluster

pihme:
My two cents on this:

* there is no Jenkins job to delete
* I would propose to keep resources re-activatable. I know we don't support 0.26 anymore, but if a customer comes, I would like to be able to reactivate everything. This means I would keep the stable branch, and I would drain the namespace, but keep the namespaces in zeebe gke (assuming an empty namespace costs no money).
* For cloud clusters we need to delete two - the productive and the dev cluster

korthout:
Please note that I've disabled testbench 0.26 for now. Due to problems with the create-cluster worker.

I've disabled it using:
```
kubectx gke_zeebe-io_europe-west1-b_zeebe-cluste
kubens testbench
kubectl scale deployment testbench --replicas 0
```

pihme:
Repeated the scale down after auto-deploy reactivated everything:
```
kubectl scale deployment testbench --replicas 0
kubectl scale deployment chaos-worker --replicas 0
```

pihme:
@npepinpe Any objections to removing 0.26 testbench? If not, I would clean this up as part of my medic duty this week

npepinpe:
No, let's remove it. The last 0.26 patch is out already, we don't need to run Testbench on this branch anymore.

pihme:
@saig0 FYI - I will delete the 0.26 clusters now

Zelldon:
Btw @pihme I dont think you need to keep the k8 namespace, since namespaces are easy to create :)

PR #444, opened 2021-07-12
chore(ci): disable deploy step

pihme:
We want to disable the 0.26 testbench auto deployment as this version is no longer
supported. This commit disables the branch name deploy trigger in order
to avoid any accidental redeployments. Manual deployments are still possible

related to #423


korthout (approved):
LGTM 👍

pihme:
bors retry

PR #449, opened 2021-07-15
chore(processes): focus on testing Production - S exclusively

pihme:
The cluster plans that were tested were based on the T-shirt sizes defined pre GA.
All new customers will be able to create clusters in a GA cluster plan which matches what Production - S was.
Therefore, we want to test this cluster plan exclusively.


processes/qa-protocol.bpmn
```diff
@@ -12,7 +12,7 @@
         <zeebe:calledElement processId="run-all-tests-in-camunda-cloud-per-cluster-plan" />
         <zeebe:ioMapping>
           <zeebe:input source="=&#34;new chaos&#34;" target="region" />
-          <zeebe:input source="=[&#34;Development&#34;, &#34;Production - S&#34;, &#34;Production - M&#34;, &#34;Production - L&#34;]" target="clusterPlans" />
+          <zeebe:input source="=[&#34;Production - S&#34;]" target="clusterPlans" />
```
korthout, line 15: Is it possible to change the `Production - S` to `GA Hardware ....`?

pihme: Yes, but this involves a couple more steps. Let's talk about it in the medic hand-over

PR #454, opened 2021-07-27
fix(core): filter for region when looking up for cluster plan

### LINKED ISSUE #211: Assumption that cluster plan name is unique is wrong
Currently we look up the cluster plan type by its name. However, there could be several cluster plan types with the same name. Each would have a different data center/region. 

Currently the information which cluster plan type belongs to which region is not available in the data.
https://github.com/camunda-cloud/console/issues/1581 will remedy that.

Once we have this information available we need to include it in the cluster plan type lookup.

deepthidevaki:
Cluster plan name is not unique; it is unique only with in a region. This commit uses the region info when looking up for cluster plan names. So when there are cluster plans with same name in different regions, it will map to the correct one.

closes #211

### DIFF
--- a/core/src/test/java/io/zeebe/clustertestbench/handler/MapNamesToUUIDHandlerTest.java
+++ b/core/src/test/java/io/zeebe/clustertestbench/handler/MapNamesToUUIDHandlerTest.java
@@ -1,0 +1,142 @@
+package io.zeebe.clustertestbench.handler;
+
+import static java.util.Map.entry;
+import static org.camunda.community.zeebe.testutils.ZeebeWorkerAssertions.assertThat;
+import static org.mockito.Mockito.when;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.zeebe.clustertestbench.cloud.CloudAPIClient;
+import io.zeebe.clustertestbench.cloud.response.ParametersResponse;
+import java.util.Map;
+import org.assertj.core.api.Assertions;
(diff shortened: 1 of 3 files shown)


core/src/test/java/io/zeebe/clustertestbench/handler/MapNamesToUUIDHandlerTest.java
```diff
@@ -0,0 +1,142 @@
...
+  ActivatedJobStub activatedJobStub;
+
+  MapNamesToUUIDsHandler sutHandler;
+
+  @BeforeEach
+  void setUp() throws JsonProcessingException {
+    sutHandler = new MapNamesToUUIDsHandler(mockCloudAPIClient);
+    activatedJobStub = jobClientStub.createActivatedJob();
+    final ParametersResponse parameterResponse =
+        new ObjectMapper().readValue(PARAMETER_RESPONSE_JSON, ParametersResponse.class);
```
pihme, line 93: :wrench: Not sure whether parsing the JSON should be part of the test. Instead, you could also just create the object graph directly

pihme (approved):
lgtm :tada: 

One optional comment

core/src/test/java/io/zeebe/clustertestbench/handler/MapNamesToUUIDHandlerTest.java
deepthidevaki: Since we don't have an integration test, I felt more confident doing this way. I took a real json response from the query and adapted it for the test. This way I know that the `ParametersResponse` is inline with the expected json response.

PR #455, opened 2021-07-29
fix: update model to timeout while waiting for the health check

### LINKED ISSUE #230: Infinite loop when cluster is deleted before it gets healthy
After creating a cluster the testbench will wait for it to become healthy. If the cluster is deleted before it got healthy, then the testbench will remain in an infinite loop.

image

deepthidevaki:
Added a timeout to the health check. If it times out the tests are not run and will be marked as failed. 

closes #230


processes/prepare-zeebe-cluster-in-camunda-cloud.bpmn
```diff
@@ -63,97 +32,195 @@
...
+      <bpmn:sequenceFlow id="condition_healthy" name="Healthy" sourceRef="Gateway_0rsc9zf" targetRef="Event_0o2ftqr">
+        <bpmn:conditionExpression xsi:type="bpmn:tFormalExpression">= clusterStatus = "Healthy"</bpmn:conditionExpression>
+      </bpmn:sequenceFlow>
+    </bpmn:subProcess>
+    <bpmn:sequenceFlow id="Flow_04hrlj4" sourceRef="create-api-client-for-cluster-in-camunda-cloud" targetRef="Activity_12o4zvy" />
+    <bpmn:sequenceFlow id="Flow_0co0hrw" sourceRef="Activity_12o4zvy" targetRef="gather-information-about-cluster-in-camunda-cloud" />
+    <bpmn:boundaryEvent id="Event_1v9wu39" attachedToRef="Activity_12o4zvy">
+      <bpmn:outgoing>Flow_08i347f</bpmn:outgoing>
+      <bpmn:timerEventDefinition id="TimerEventDefinition_1e4cu8f">
+        <bpmn:timeDuration xsi:type="bpmn:tFormalExpression">PT30M</bpmn:timeDuration>
```
pihme, line 89: * please increase the timeout to 2h (we already had instances where it took quite some time for a cluster to become healthy)
* Please also give a name to the timer event. The name will be visible in the BPMN diagram

pihme (changes requested):
lgtm

Please look at the one comment I had.
Please also update the screenshots of the processes in `docs/assets/

pihme (approved):
Please also update the screenshot for the second process that was changed

PR #458, opened 2021-08-03
285 reenable integration test

### LINKED ISSUE #285: Re-enable TriggerMessageStartEventHandlerIT
This test was disabled, because it uses ZeeQS, which at the time of #249 is not yet compatible with Zeebe 1.0.0 versions.

pihme:
This change re-enables the integration test, which was disabled due to version incompatibilities,

closes #285


deepthidevaki (approved):
Thanks :+1:

bors r+

PR #459, opened 2021-08-03
Check whether cluster still exists

### LINKED ISSUE #223: Delete Cluster worker should be idempotent
I encountered an incident due to 404 on DELETE cluster. 

The incident in the workflow instance
Screenshot 2021-01-25 at 14 28 10

The incident details:
Screenshot 2021-01-25 at 14 29 20

I think this did not have to be an incident. When the worker tries to delete the cluster and it is not there, then this is actually the same as it's intended result. 

**Expected behaviour**
The worker should complete the job successfully when receiving 404 on DELETE cluster.

pihme:
The change adds a check whether the cluster still exists before it is deleted.

closes #223


deepthidevaki (approved):
:+1:

bors r+

PR #460, opened 2021-08-03
feat(core): add cluster id to slack message

### LINKED ISSUE #345: Add cluster UUID to slack message
Currently the slack message for a failure only contains the cluster name. Often we need the UUID for the cluster. Putting this UUID directly into the message would save a lookup.

pihme:
This change adds the cluster Id for easier reference to the slack message.

closes #345


deepthidevaki (approved):
:+1:

bors r+

PR #483, opened 2021-10-04
fix(chaos): avoid getting stuck on run command

saig0:
* redirecting the error stream to avoid getting stuck if it is written to the error stream
* setting a timeout to make it more visible if the process is stuck


Zelldon (approved):
Thanks lets try :sweat_smile:

Zelldon:
It seems ci doesnt work @saig0

PR #485, opened 2021-10-07
Enable dependabot to update github actions

korthout:
This should configure dependabot to keep track of updates for github actions. Currently, that is only the backport-action.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,8 @@ updates:
     schedule:
       interval: "daily"
     target-branch: "develop"
+  - package-ecosysten: "github-actions"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


pihme (approved):
bors merge

.github/dependabot.yml
```diff
@@ -10,3 +10,8 @@ updates:
     schedule:
       interval: "daily"
     target-branch: "develop"
+  - package-ecosysten: "github-actions"
```
npepinpe, line 13: Is a typo? :scream: 

`package-ecosysten` should be `package-ecosystem` right?

korthout: Yup, I fixed in 81a83d565f679aacfbbba74256c424fbd184b977 😅

PR #512, opened 2021-11-30
Migrate chaos workers to chaos repo

Zelldon:
I think we discussed it once that at some point it might make sense to migrate the workers, so the time has come.

Today I moved the chaos workers out to the zeebe-chaos repository and released a new version so we can already use it in testbench https://github.com/zeebe-io/zeebe-chaos/releases/tag/1.0.0. 

Currently the workers are still using the latest master scripts,  I will fix this soon as well so we only use the released versions, which makes it easier to iterate over scripts and add scripts without breaking current builds etc. Furthermore I plan to migrate more scripts to kotlin workers, which should helps us in maintaining it. Since the chaos experiments and everything what is chaos related was located in the zeebe-chaos repo I think it makes sense that we move the workers also to that repo. I hope this change is also ok for you.

I tried to adjust the pipeline and one of the readme. I proably have missed some more.


pihme (approved):
LGTM :tada:

PR #517, opened 2021-12-03
ci: bump chaos worker version

Zelldon:
Key feature: Everything is bundled in the image and jar, which means we can now change zeebe-chaos master without causing issues.


Zelldon:
I found an issue, I will reopen it later

PR #518, opened 2021-12-03
ci: update zeebe-chaos

Zelldon:
**Key feature:** Everything is bundled in the image and jar, which means we can now change zeebe-chaos master without causing issues.

Release 1.2.0 https://github.com/zeebe-io/zeebe-chaos/releases/tag/1.2.0
Release 1.1.0 https://github.com/zeebe-io/zeebe-chaos/releases/tag/1.1.0

Note: Ideally we could update the used version via dependabot but this is currently not supported.

### DIFF
--- a/.ci/scripts/deploy.sh
+++ b/.ci/scripts/deploy.sh
@@ -9,7 +9,7 @@ then
 fi
 
 tag=$1
-workerVersion=1.0.0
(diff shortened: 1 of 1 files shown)


pihme (approved):

pihme:
bors merge

PR #569, opened 2022-02-21
fix: finish deploy when there are shutdown pods

oleschoenburg:
Waiting for condition ready does not work if there are pods that are shut down.
Since we are restarting a rollout before, we can just wait for the rollout to complete


deepthidevaki (approved):
Thanks!

PR #591, opened 2022-03-07
chore(it): refactor test to use zeebe-process-test instead of

pihme:
- remove dependency to BPMN Spec
- rewrite test to use zeebe-process-test

### DIFF
--- a/integration-tests/pom.xml
+++ b/integration-tests/pom.xml
@@ -46,12 +46,9 @@
       <scope>test</scope>
     </dependency>
 
-    <!-- running the spec with Zeebe -->
     <dependency>
-      <groupId>io.zeebe.bpmn-spec</groupId>
-      <artifactId>zeebe-test-runner</artifactId>
-      <version>${bpmn-spec.version}</version>
-      <scope>test</scope>
+      <groupId>io.camunda</groupId>
+      <artifactId>zeebe-process-test-extension-testcontainer</artifactId>
     </dependency>
   </dependencies>
 
--- a/integration-tests/src/test/java/io/zeebe/clustertestbench/it/handler/TriggerMessageStartEventHandlerIT.java
+++ b/integration-tests/src/test/java/io/zeebe/clustertestbench/it/handler/TriggerMessageStartEventHandlerIT.java
@@ -1,39 +1,32 @@
 package io.zeebe.clustertestbench.it.handler;
 
-import static java.util.stream.Collectors.toMap;
+import static io.camunda.zeebe.process.test.assertions.BpmnAssert.assertThat;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.awaitility.Awaitility.await;
 
 import io.camunda.zeebe.client.ZeebeClient;
+import io.camunda.zeebe.client.api.response.DeploymentEvent;
 import io.camunda.zeebe.client.api.response.ProcessInstanceEvent;
 import io.camunda.zeebe.client.api.worker.JobWorker;
-import io.zeebe.bpmnspec.runner.zeebe.ZeebeEnvironment;
-import io.zeebe.bpmnspec.runner.zeebe.zeeqs.ZeeqsClient;
-import io.zeebe.bpmnspec.runner.zeebe.zeeqs.ZeeqsClient.ElementInstanceDto;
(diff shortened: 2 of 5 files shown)


integration-tests/src/test/java/io/zeebe/clustertestbench/it/handler/TriggerMessageStartEventHandlerIT.java
```diff
@@ -80,32 +73,22 @@ void shoulTriggerSecondaryProcess() {
...
-    // then
-    await()
-        .atMost(1, TimeUnit.MINUTES)
-        .until(() -> zeeqsClient.getProcessInstanceKeys().size() == 2);
-
-    final List<Long> processInstanceKeys = zeeqsClient.getProcessInstanceKeys();
-
-    processInstanceKeys.remove(mainProcessCreationResponse.getProcessInstanceKey());
+    // TODO replace with waitForIdleState
+    Thread.sleep(250);
```
pihme, line 77: :warning: Interesting detail is that if I replace this line with `engine.wait()` I will get test failures.
@remcowesterhoud Is this expected, or should I create a bug for this?

integration-tests/pom.xml
```diff
@@ -46,12 +46,9 @@
...
     </dependency>
 
-    <!-- running the spec with Zeebe -->
     <dependency>
-      <groupId>io.zeebe.bpmn-spec</groupId>
-      <artifactId>zeebe-test-runner</artifactId>
-      <version>${bpmn-spec.version}</version>
-      <scope>test</scope>
+      <groupId>io.camunda</groupId>
+      <artifactId>zeebe-process-test-extension-testcontainer</artifactId>
```
remcowesterhoud, line 51: 🔧 The scope could be changed to `test`

integration-tests/src/test/java/io/zeebe/clustertestbench/it/handler/TriggerMessageStartEventHandlerIT.java
remcowesterhoud: Please do. It seems it doesn't work properly when combined with job workers. We actually have the same issue in the `WorkerTest` in ZPT, where I'm also using a `Thread.sleep()` to work around it.

I should've created an issue at that time, because I forgot about it 😅

remcowesterhoud (commented):
Awesome! I have just 1 minor suggestion, the rest LGTM if you actually want to merge it, since it is an alpha version and next version will contain breaking changes.

remcowesterhoud (approved):

integration-tests/src/test/java/io/zeebe/clustertestbench/it/handler/TriggerMessageStartEventHandlerIT.java
```diff
@@ -80,32 +73,22 @@ void shoulTriggerSecondaryProcess() {
...
-        variablesInSecondaryProcess.stream()
-            .collect(
-                toMap(
-                    VariableDto::getName,
-                    dto -> dto.getValue().replace("\"", ""))); // value is wrapped in " characters
+    final InspectedProcessInstance lastProcessInstance = optProcessInstance.get();
 
-    assertThat(variablesInSecondaryProcessAsMap).containsExactlyInAnyOrderEntriesOf(variables);
+    assertThat(lastProcessInstance).hasVariableWithValue("key1", "value1");
+    assertThat(lastProcessInstance).hasVariableWithValue("key2", "value2");
```
remcowesterhoud, line 92: 💭 Maybe an assertion in ZPT to assert a map of variable could be valuable.

integration-tests/pom.xml
pihme: That's already done in parent pom. I assume the scope gets inherited as well

integration-tests/src/test/java/io/zeebe/clustertestbench/it/handler/TriggerMessageStartEventHandlerIT.java
pihme: https://github.com/camunda-cloud/zeebe-process-test/issues/243

pihme:
bors merge

PR #593, opened 2022-03-08
582 spotless

pihme:
Adds spotless plugin


pihme:
@korthout Could use your expertise here. It runs on my machine, but on Jenkins it finds violations. Not sure what to make of this. (I also tried removing the MAVEN_SETTINGS in the Maven call, but it didn't change anything)

pihme:
bors merge

PR #594, opened 2022-03-08
Upgrade to Java 17 & Use In-Memory Engine for tests

### LINKED ISSUE #586: Bump to Java 17
* [ ] Use Java 17
* [ ] Also change zeebe process test extension to use the in memory one

oleschoenburg:
closes #586

### DIFF
--- a/.ci/podSpecs/distribution.yml
+++ b/.ci/podSpecs/distribution.yml
@@ -13,7 +13,7 @@ spec:
       emptyDir: {}
   containers:
     - name: maven
-      image: maven:3.6.3-openjdk-11
+      image: maven:3-eclipse-temurin-17
       command: ["cat"]
       tty: true
       env:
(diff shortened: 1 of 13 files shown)


pihme (changes requested):
All the changes so far look good.

Two changes are still missing:
* `lgtm.yml` 
* `Dockerfile`

Then, we should do a test deployment to the dev environment. On Jenkins you can select _Run with parameters_ and check _Deploy to Dev_  (leaving all other fields unchanged. Unfortunately, this all works and looks ok.

But then if you look at K8S namespace _testbench-1-x-dev_ in _zeebe-cluster_ namespace you can see that the testbench is not starting.

```
NAME                            READY   STATUS             RESTARTS   AGE
chaos-worker-7b5795f59f-bqh9c   1/1     Running            0          6m28s
testbench-6f88887c4d-rblcg      0/1     CrashLoopBackOff   6          6m59s
```
I don't want to exploit your kindness. So if this is getting too long winded,I am happy to take over. But I always use every chance to expose more colleagues to the testbench so that this project gets more brainspace.

pihme (approved):
Thanks a lot!

PR #596, opened 2022-03-09
Spring Cleaning Part I

pihme:
- use Java 17 features (mostly records)
- code cleanup

Changes are focused on `cloud-client` module to keep the PR manageable.


pihme:
@korthout If you want to take a sneak peak, maybe have a look at the commits regarding records. I would like to know your opinion on putting records into interfaces.

korthout:
@pihme I love it! The interfaces specify clearly what data they require. Records are awesome for this. I'm a fan 👯

pihme:
@korthout Interesting. I thought you might object to it. Implementing records in the interface felt as wrong to me as implementing the first default methods, when these became a feature. At the same time, it makes so much sense.

So thanks for your approval. That makes me feel much better, and I will go ahead and find more records.

pihme:
@korthout Would be great if you could review this as medic. If you don't get to it, maybe you could hand it over to the next medic?

Please ignore LGTM which does not support Java 17 and will be removed. 

Best way to review this is commit by commit. The changes are straightforward, which increases the danger for me making stupid mistakes. And we don't have too many tests in that area, unfortunately.

core/src/main/java/io/zeebe/clustertestbench/handler/CheckGenerationUsageHandler.java
```diff
@@ -42,15 +42,15 @@ public String getGenerationUUID() {
...
     }
   }
 
   static class Output {
     private final boolean generationNotInUse;
 
-    public Output(boolean generationNotInUse) {
+    public Output(final boolean generationNotInUse) {
       this.generationNotInUse = generationNotInUse;
     }
```
korthout, line 56: ❓ Can the zeebe client deal with records as variables? It might be interesting to give it a try to convert `Input` and `Output` to records and see if that works.

korthout (approved):
Already LGTM 👍  I do have a few suggestions, so please have a look.

core/src/main/java/io/zeebe/clustertestbench/handler/CheckGenerationUsageHandler.java
pihme: Will try this in Spring cleaning 2 or 3 :-)

PR #363, opened 2021-05-19
Improve chaos logging

Zelldon:
1. In order to log via json layout, which is better visible and processed
by stackdriver, this commits adds a dependency to the zeebe-util module
and uses the StackDriverLayout
2. Use MDC for logging, add as context information:

 * the job type
 * cluster id
 * cluster plan
 * process instance id
3. Furthermore the process logs are redirected to log4j and uses the
correct context

Example output:

```
{"severity":"DEBUG","logging.googleapis.com/sourceLocation":{"function":"invoke","file":"chaosMain.kt","line":159},"message":"+ kubectl wait --for=condition=Ready pod -l app.kubernetes.io/app=zeebe --timeout=900s -n 0ee5902e-1dfa-4805-9eb2-51d8036faaa2-zeebe","serviceContext":{"service":"chaos-worker","version":"development"},"context":{"clusterPlan":"Production - M","threadId":25,"processInstanceId":"4503599627372936","clusterId":"0ee5902e-1dfa-4805-9eb2-51d8036faaa2","jobType":"verify-readiness.sh","threadPriority":5,"loggerName":"io.zeebe.chaos.ChaosWorker","threadName":"pool-3-thread-1"},"timestampSeconds":1621427016,"timestampNanos":690601000}
{"severity":"DEBUG","logging.googleapis.com/sourceLocation":{"function":"invoke","file":"chaosMain.kt","line":159},"message":"error: no matching resources found","serviceContext":{"service":"chaos-worker","version":"development"},"context":{"clusterPlan":"Production - M","threadId":25,"processInstanceId":"4503599627372936","clusterId":"0ee5902e-1dfa-4805-9eb2-51d8036faaa2","jobType":"verify-readiness.sh","threadPriority":5,"loggerName":"io.zeebe.chaos.ChaosWorker","threadName":"pool-3-thread-1"},"timestampSeconds":1621427016,"timestampNanos":691035000}
{"severity":"INFO","logging.googleapis.com/sourceLocation":{"function":"run","file":"chaosMain.kt","line":87},"message":"Received shutdown signal","serviceContext":{"service":"chaos-worker","version":"development"},"context":{"threadId":33,"threadPriority":5,"loggerName":"io.zeebe.chaos.ChaosWorker","threadName":"Close thread"},"timestampSeconds":1621427054,"timestampNanos":479319000}
```

Interesting is the context part:

```
[...]

PR #597, opened 2022-03-09
Improve testbench worker logging

### LINKED ISSUE #344: Improve log messages for workers
Currently all workers write to the same log stream. This makes it difficult to filter for e.g. all log messages related to a certain worker and related to a certain cluster.

One improvement would be to use MDC https://dzone.com/articles/mdc-better-way-of-logging-1 to also log metadata
* which worker produced the log
* Which cluster is being tested
* Which process instance (in testbench) triggered the worker

We could then make this information available to StackDriver, similarly to what we do in Zeebe.

korthout:
Improves the testbench logging in 2 ways:
- Applies stackdriver layout to logging when deployed to dev and prod.
- Enhances the logging with job-related data
- Switches from assembly plugin to shade plugin

These changes have been inspired by https://github.com/zeebe-io/zeebe-cluster-testbench/pull/363

closes #344


korthout:
@pihme I'm not sure anymore how to test this on dev. Any help would be appreciated 🙇

core/src/main/java/io/zeebe/clustertestbench/util/LogDetails.java
```diff
@@ -0,0 +1,47 @@
+package io.zeebe.clustertestbench.util;
+
+import io.camunda.zeebe.client.api.response.ActivatedJob;
+import java.util.Map;
+import org.slf4j.MDC;
+
+public final class LogDetails {
```
pihme, line 7: :x: I think this should create an auto-closable context which calls `MOC.clear()` in its `close()` method. This now writes into a log context on the thread, but doesn't remove those details. So any subsequent call on the same thread would use the same details. If the next job worker also calls `LogDetails.setMOCForJob(job)` it might work. But if the next job handler has less variables, it will not clear the variables it doesn't have.

If, however, this class would create a log context, then the job handler could become:
```
try (final logContext = LogDetails.forJob(job)) {
  ...
}
```

And once we have this general structure, we can also make this a decorator around a job handler. And add this decorator in `io.zeebe.clustertestbench.bootstrap.Launcher#registerWorker`

pihme, line 7: :x: Please also add a test for this class (e.g. after `setMOCForJob(job)` is called, then `MOC` has certain contents.

`ActivatedJobStub` might be helpful here

core/src/main/resources/log4j2.xml
```diff
@@ -1,17 +1,29 @@
...
+    </Console>
+
+    <Console name="Stackdriver" target="SYSTEM_OUT">
+      <StackdriverLayout serviceName="${log.stackdriver.serviceName}"
+        serviceVersion="${log.stackdriver.serviceVersion}"/>
     </Console>
   </Appenders>
+
   <Loggers>
+    <Logger name="io.zeebe.clustertestbench.cloud.filter" level="debug" additivity="false"/>
```
pihme, line 20: :x: I think this produces too much noise. Some requests are small, but some request produce KB worth of log output

testbench-prod.yaml
```diff
@@ -71,6 +71,12 @@ spec:
                 secretKeyRef:
                   name: testbench-secrets
                   key: slackWebhookUrl
+            - name: ZCTB_LOG_APPENDER
+              value: "Stackdriver"
+            - name: ZCTB_LOG_STACKDRIVER_SERVICEVERSION
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
```
pihme, line 79: :question: Can you please explain how/why this works? I am curious to learn more about this.

pihme (changes requested):
Good start. Some suggestions for improvements.

testbench-prod.yaml
korthout: Sure! What this does is set the value of the `ZCTB_LOG_STACKDRIVER_SERVICEVERSION` env var to the k8s namespace that this is deployed to. 

For example, we currently deploy to the namespaces `testbench-1-x-prod` and `testbench-1-x-dev`. That namespace is then used and we see that in the log explorer under `jsonPayload.serviceContext.version`. 

This same version value is used in the Error Reporting Tool for filtering. We already use this field in the same way in the benchmark project.

core/src/main/resources/log4j2.xml
korthout: Yes, perhaps it's not necessary all the time. Note that it is currently configured like this as well

testbench-prod.yaml
pihme: Ah, so it's not the version of StackDriver. It is the version that appears as metadata in log items. Got it! :bow:

core/src/main/resources/log4j2.xml
pihme: Well, then this is a good moment to change it.

core/src/main/java/io/zeebe/clustertestbench/handler/JobHandlerWithEnrichedLogger.java
```diff
@@ -0,0 +1,89 @@
...
+import io.camunda.zeebe.client.api.worker.JobHandler;
+import java.util.List;
+import java.util.Map;
+import org.slf4j.MDC;
+
+/**
+ * Job handler decorator that enriches the logging framework with context information from the
+ * activated job.
+ */
+public final class JobHandlerWithEnrichedLogger implements JobHandler {
```
pihme, line 14: @deepthidevaki Might also be interesting for your e2e tests

deepthidevaki: Thanks. Looks like a good idea that we can adopt into e2e tests.

korthout:
@pihme I've applied your hints, and manually tested it. Please have another look

core/src/main/java/io/zeebe/clustertestbench/handler/JobHandlerWithEnrichedLogger.java
```diff
@@ -0,0 +1,89 @@
...
+
+  /**
+   * This logging enricher uses MDC to provide context information to the logging framework. It's an
+   * autoclosable to make sure the MDC is cleaned-up again after usage.
+   */
+  public static final class LoggingEnricher implements AutoCloseable {
+
+    private static final String JOB_TYPE = "jobType";
+    private static final String JOB_PROCESS_INSTANCE_KEY = "processInstanceKey";
+    private static final List<String> JOB_VARIABLES =
```
pihme, line 38: :heart_eyes: Love it!

core/src/main/java/io/zeebe/clustertestbench/handler/JobHandlerWithEnrichedLogger.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private LoggingEnricher(final ActivatedJob job) {
+      putJobValuesInMDC(job);
+    }
+
+    @Override
+    public void close() {
+      MDC.remove(JOB_TYPE);
+      MDC.remove(JOB_PROCESS_INSTANCE_KEY);
+      JOB_VARIABLES.forEach(MDC::remove);
```
pihme, line 60: :heart_eyes: Love it!

core/src/test/java/io/zeebe/clustertestbench/handler/JobHandlerWithEnrichedLoggerTest.java
```diff
@@ -0,0 +1,94 @@
...
+  @Test
+  void shouldEnrichMDCWhenJobIsBeingHandled() throws Exception {
+    // given
+    final var job = DataHelper.createActivatedJob();
+
+    // when
+    new JobHandlerWithEnrichedLogger(
+            (c, j) ->
+                // then
+                assertThat(MDC.getCopyOfContextMap())
```
pihme, line 38: :wrench: I think this test would be more readable if it only tested the enriched logger in isolation

pihme (approved):
Thanks so much for putting in the time and effort (especially the logging and assembly and shading bit). 

:heart_eyes: I love this a lot! :heart_eyes:

core/src/test/java/io/zeebe/clustertestbench/handler/JobHandlerWithEnrichedLoggerTest.java
korthout: Yeah, I thought about this. But wasn't sure how to improve things, so I just added a test that verified that the delegation works. If you have ideas on how to improve this quickly I'd be willing to look into it, but I don't want to spend too much time on this anymore 😄

pihme: Pseudo code:
```
//when 
loggingEnricher = new LoggingEnricher(job)
map1 = MDC.getCopyOfContextMap())
loggingEnricher.close()
map2 = MDC.getCopyOfContextMap())

//then
assertThat(map1) ...
assertThat(map2).isEmpty()

```

korthout: @pihme I've given it a quick try, let me know what you think.

pihme: Looks good to me. To be safe, I would suggest to also call close in the first test.

korthout:
bors merge

korthout:
Seems like an unrelated build failure (some network connection issue). Let's retry that

bors merge

PR #600, opened 2022-03-11
Remove chaos files

pihme:
This PR removes the chaos files. 

Most of the files have already been moved to chaos test. However, some files are duplicated which leads to confusion and possibly to unintended outcomes: https://github.com/zeebe-io/zeebe-chaos/issues/123

With the upcoming changes (i.e. Deepthi implementing another test) this repo will be slimmed down to only provide the bare testbench, which is then extensible by independently deployed tests (chaos tests, the Deepthi test).

@Zelldon Happy to discuss this direction of travel with you, if you have any objections. This is not set in stone. Since you were off this week, I didn't yet have a chance to align. This PR was created under the assumption that you agree to this approach.

@Zelldon I checked whether all the stuff I removed here is available in chaos-test. This seems true for the BPMN models and the deployment descriptors. However, I didn't immediately find any scripts to actually deploy the chaos workers. So this logic might have to be moved, or we just drop the last commit for a time being.


Zelldon:
Please sign the CLA @pihme  :laughing: (never thought that I would say that to you :laughing: )

pihme:
Superseeded by #632  and then a new PR for the last commit - at the appropriate time

Zelldon (approved):

Zelldon:
Merging it since we don't want to redeploy the old chaos workers anymore. We migrated to zbchaos v1.0.0 https://github.com/zeebe-io/zeebe-chaos/issues/139

We will add a deploy pipeline in zeebe-chaos repo soon.

PR #603, opened 2022-03-11
Maven pom spring cleaning

korthout:
This cleans up the pom files:
- improves the readability by applying a consistent style
- moves all version definitions into dependency management and plugin management
- moves all properties into the root project
- organizes dependencies into a few main categories
- ~re-enables the maven enforcer plugin, reduces the strictness from `dependencyConvergence` to `requireUpperBoundDeps`~
- removes unnecessary dependencies on `zeebe-util` and `zeebe-test-util` modules

This pull request had been on hold since it was first opened in March 2022. This was at the time that ZPT hadn't been officially released yet, and this project depended on a SNAPSHOT version of it. Dependency convergence between ZPT and the Zeebe Client was a major reason that I was unable to re-enable the maven enforcer plugin. I'm not sure that we should block this pull request from being merged any longer for that reason. If we want to re-enable the maven enforcer plugin, then we also do that in a separate pull request.

### DIFF
--- a/testdriver/api/src/test/java/io/zeebe/clustertestbench/testdriver/api/JsonAssertions.java
+++ b/testdriver/api/src/test/java/io/zeebe/clustertestbench/testdriver/api/JsonAssertions.java
@@ -1,0 +1,20 @@
+package io.zeebe.clustertestbench.testdriver.api;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.assertj.core.api.Assertions;
+
+public class JsonAssertions {
+
+  public static void assertJsonEquality(final String actual, final String expected) {
+    final var objectMapper = new ObjectMapper();
+    try {
+      Assertions.assertThat(objectMapper.readTree(actual))
+          .isEqualTo(objectMapper.readTree(expected));
+    } catch (final JsonProcessingException e) {
+      Assertions.fail(
+          "Expected valid JSON, but 'actual' was '%s' and 'expected' was '%s'"
+              .formatted(actual, expected));
+    }
+  }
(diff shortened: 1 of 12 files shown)


testdriver/api/src/test/java/io/zeebe/clustertestbench/testdriver/api/JsonAssertions.java
```diff
@@ -0,0 +1,20 @@
+package io.zeebe.clustertestbench.testdriver.api;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.assertj.core.api.Assertions;
+
+public class JsonAssertions {
+
+  public static void assertJsonEquality(final String actual, final String expected) {
```
Zelldon, line 9: Why is this now duplicated?

Zelldon (approved):
Thanks @korthout :+1: looks good, small question did we use sort pom now? As we do in zeebe?

testdriver/api/src/test/java/io/zeebe/clustertestbench/testdriver/api/JsonAssertions.java
korthout: Previously this code came from the dependency on `zeebe-test-util`. I did not find a place in this repo to put this code just once. But I don't expect these few lines of code ever to change, so IMO this duplication shouldn't matter 🤷 

If you feel otherwise, feel free to duplicate 👍

korthout:
>did we use sort pom now?

@Zelldon This PR does not change the used plugins for styling/formatting. This repo already used sortpom via spotless before.

korthout:
bors merge

PR #609, opened 2022-03-17
test(cloud-client): add test for marshalling roundtrip in RestEasy

pihme:
This test uses wiremock to set up an HTTP server for the unit test.
With this test server it is possible to test a request/response roundtrip
that uses RestEasy marshalling. This is more realistic than using Jackon
directly.

Also while working on this test, I realized that the OAuthCredentials
object could be simplified and did so. In doing so the field "expiry"
was removed. This field was not used in the code. Basically, we try
to use the token, and if it has expired, we will request a new token.


cloud-client/src/test/java/io/zeebe/clustertestbench/cloud/oauth/OAuthClientTest.java
```diff
@@ -59,4 +76,86 @@ void testSerializationForUserAccountTokenRequest() throws JsonProcessingExceptio
...
+    final OAuthCredentials actual =
+        OBJECT_MAPPER.readValue(exampleResponse, OAuthCredentials.class);
+
+    // then
+    assertThat(actual.tokenType()).isEqualTo("Bearer");
+    assertThat(actual.accessToken()).isEqualTo("lorem ipsum");
+  }
+
+  @Test
+  void testRoundTripForUserAccountTokenRequestAsPartOfRestEasyMarshalling(
```
oleschoenburg, line 101: :upside_down_face:  I think this test name has become a bit too verbose. The jackson tests are testing serialization, and the RestEasy tests are testing the roundtrip.

oleschoenburg (approved):
LGTM :+1: I like how LOC are shrinking and test coverage is increasing.

cloud-client/src/test/java/io/zeebe/clustertestbench/cloud/oauth/OAuthClientTest.java
pihme: yolo

pihme:
bors abort

pihme:
bors cancel

pihme:
bors retry

PR #613, opened 2022-03-21
cleanup(processes): remove obsolete process

### LINKED ISSUE #612: Remove process run-all-tests-in-camunda-cloud-per-region
The process is not being used

pihme:
closes #612


deepthidevaki (approved):
region and regionUUID is used in several places.  Is there a default region if we are not running tests in all regions?

PR #615, opened 2022-03-21
Reduce `TestReport` interface

### LINKED ISSUE #614: Reduce `TestReport` interface
New minimal test report:
```
public interface TestReport {

  TestResult getTestResult();

  List<String> getFailureMessages();

  public enum TestResult {
    PASSED,
    FAILED,
    SKIPPED;    
  }
}
```

pihme:
Several commits to reduce the `TestReport` interface and clean up the code a little bit to carve out which is the main interface for a test

closes #614 

**Review Hints**
- probably best review approach is commit by commit
- there is a bit of inconsistency in when I use a record for something and when an interface. Not fully satisfied with this myself. The main idea was if it is something, I might want to extend from, it's an interface. This often happens to be output types. On the other hand, if it is something that is given, I use a record. This often happens to be input types.
- please don't merge; I want to test the changes on Dev before merging
- regarding the `TestDriver` interface I am still unsure what is helpful in terms of reuse. I think it's a step forward. I also thought about implementing a job handler that delegates to a test driver and takes care of variables and exception handling. But I wasn't sure this is really helpful. Both Chris and you have implemented a whole suite or processes for a test. Maybe when you try to use it you get an idea of which bits could be shared.

/cc @Zelldon


docs/extension-guide.md
```diff
@@ -9,18 +9,27 @@ Additional tests can be added very easily by:
...
+The called process will find the variables that are defined in `TestInputDTO`. This includes _authenticationDetails_ which
 can be used to authenticate against the Zeebe cluster.
 
-The called process is expected to return a _testReport_ (see class `TestReport` for the expected structure)
+The called process is expected to return an output - see class `TestOutput` for the expected structure.
+
+The called process can have sub processes and complex structures. In its simples form, the called process can have a single service task
+that delegates to a `TestDriver`. A test driver is responsible for running a test within a cluster provided by Testbench.
+
+Please have a look at `TestDriver` which defines the interface of a test driver.
```
Zelldon, line 28: :wrench: A link would be good here

docs/extension-guide.md
```diff
@@ -9,18 +9,27 @@ Additional tests can be added very easily by:
...
 ![Run Test In Camunda Cloud](assets/run-test-in-camunda-cloud.png)
 
 This process takes as parameters, among others, the _testProcessId_. This is the ID of the process you want to have called.
 
-The called process will find in the variables, among others, _authenticationDetails_ which
+The called process will find the variables that are defined in `TestInputDTO`. This includes _authenticationDetails_ which
 can be used to authenticate against the Zeebe cluster.
 
-The called process is expected to return a _testReport_ (see class `TestReport` for the expected structure)
+The called process is expected to return an output - see class `TestOutput` for the expected structure.
```
Zelldon, line 23: :question: Does it mean we changed the structure? Do we need to adjust the chaos worker?

pihme: Yes, we changed the structure. However, the changes amount to a reduction (i.e. no more timestamps) so the existing output should be compatible with the reduced structure (still need to do a test run, though). 

So no changes to chaos tests are needed. At the same time, you can now ditch the start time/end time stuff, if you want. These were needed by the step to export records to Google Sheet, which has been eliminated.

Zelldon: Thanks @pihme :+1: :bow:

testdriver/api/src/main/java/io/zeebe/clustertestbench/testdriver/api/CamundaCloudAuthenticationDetails.java
```diff
@@ -1,16 +1,31 @@
 package io.zeebe.clustertestbench.testdriver.api;
 
-public interface CamundaCloudAuthenticationDetails {
+import io.camunda.zeebe.client.impl.oauth.OAuthCredentialsProvider;
+import io.camunda.zeebe.client.impl.oauth.OAuthCredentialsProviderBuilder;
 
-  String VARIABLE_KEY = "authenticationDetails";
+public record CamundaCloudAuthenticationDetails(
```
deepthidevaki: :+1:

testdriver/api/src/main/java/io/zeebe/clustertestbench/testdriver/api/TestDriver.java
```diff
@@ -1,9 +1,100 @@
...
+      }
+    }
+  }
+
+  /** Minimal implementation of {@code TestOutput} */
+  record TestOutputDTO(TestReport testReport) implements TestOutput {}
+
+  /** Minimal implementation of {@code TestReport} */
+  record TestReportDTO(TestResult testResult, int failureCount, List<String> failureMessages)
+      implements TestReport {}
```
deepthidevaki, line 69: :wrench: Not sure if they belong to this interface as they are implementations with additional fields. As an API we decided to only have `testResult` and `failureMessages`. This will be what the other tests (chaos test, e2e test) provides as output. So having this record here could be confusing because it has additional non-optional fields.

deepthidevaki (commented):
Looks good to me. Only some minor comments before I approve.

Regarding `TestDriver` interface - I'm only using the `TestInput` and `TestOutput` as the api specs between the e2e tests and testbench. I'm not using the `runTest` interface, not sure if that is helpful at all. I assume it is same for chaos tests. So as long as the TestInput and TestOuput interface is properly defined I'm happy. Bonus point, if I can import it into e2e tests so that I don't have to copy paste the interface.

testdriver/api/src/main/java/io/zeebe/clustertestbench/testdriver/api/TestDriver.java
deepthidevaki: Then it should be part of the interface, right?

pihme: But it is already :question:

deepthidevaki: Yes. You are right. I was looking at the first commit, where the new interface was introduced. 

You can ignore this comment :)

deepthidevaki (approved):

pihme:
bors merge

PR #619, opened 2022-03-24
feat: flatten all POMs for deployment

### LINKED ISSUE #580: Add Flatten POM

remcowesterhoud:
Introduces the `flatten-maven-plugin` in the build process. The projects are flattened according to the `ossrh` rules.

I took a shortcut with the description, by setting it to the name of the module. Before we didn't have any descriptions at all and this would take preference over the description that we would otherwise inherit from the `camunda-release-parent` pom.

closes #580

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@ config.properties
 /secrets-dev.yaml
 **/.checkstyle
 **/.factorypath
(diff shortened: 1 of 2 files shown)


pihme:
bors merge

PR #623, opened 2022-03-28
ci: open java modules for nexus-staging-maven-plugin

pihme:
Copy and paste of https://github.com/camunda/zeebe/pull/8130 

> Adding add-opens is a workaround for [OSSRH-66257](https://issues.sonatype.org/browse/OSSRH-66257) and [NEXUS-26993](https://issues.sonatype.org/browse/NEXUS-26993). Because we don't control the plugin ourselves and instead inherit it from camunda-release-parent we can't just add a dependency overwrite
as suggested in [OSSRH-66257](https://issues.sonatype.org/browse/OSSRH-66257).


oleschoenburg (approved):
Makes sense to me :+1:

pihme:
bors merge

PR #632, opened 2022-04-01
Remove chaos chaos keep deployment

pihme:
Removes most of the chaos files, but keeps the one needed to deploy chaos workers


Zelldon (commented):
I think we have on issue :sweat_smile: https://github.com/zeebe-io/zeebe-cluster-testbench/blob/main/.ci/scripts/deploy.sh#L56-L58 here are the deployments still used. Let me see whether I can find the time next week to adjust the chaos deployment then we can merge this

pihme:
Won't be done

PR #633, opened 2022-04-01
chore: use log level debug by default

pihme:
Previously, log level was filtered to INFO or higher. This was when
we looked at the log in K8S console and the debug logs were very verbose.

Recently, we switched to stackdriver logs. With stackdriver logs looking
at the logs in console is no longer convenient. Instead, we now use StackDriver
which also gives us the option to filter on the log level.


korthout (approved):
Makes sense to me 👍

pihme:
bors merge

PR #634, opened 2022-04-01
Allow multiple concurrent invocations

### LINKED ISSUE #630: Only one notification is sent if multiple test runs are failing
See releated slack thread https://camunda.slack.com/archives/C01JR6NLXB4/p1648709059175949

> Deepthi I guess because we use here a message start event https://bru-3.operate.ultrawombat.com/7b2cd3f5-cfcc-4a65-ab22-0ccaa071e68e/instances?process=analyse-cluster&version=1&active=true&incidents=true and only one instance can run.
> We use here https://github.com/zeebe-io/zeebe-cluster-testbench/blob/9daf4011ccd42a89139dc9948[…]e/clustertestbench/handler/TriggerMessageStartEventHandler.java the same correlation key.
>
> **See the docs where it states:**
> https://docs.camunda.io/docs/components/modeler/bpmn/message-events/
>
>  **The correlationKey of a published message can be used to control the process instance creation. If an instance of this process is active (independently from its version) and it was triggered by a message with the same correlationKey, the message is not correlated and no new instance is created.**

pihme:
Allows multiple concurrent invocations of the triggered process by using an empty correlation key.

closes #630

### DIFF
--- a/core/src/main/java/io/zeebe/clustertestbench/handler/TriggerMessageStartEventHandler.java
+++ b/core/src/main/java/io/zeebe/clustertestbench/handler/TriggerMessageStartEventHandler.java
@@ -46,7 +46,7 @@ public class TriggerMessageStartEventHandler implements JobHandler {
     zeebeClient
         .newPublishMessageCommand()
         .messageName(messageName)
-        .correlationKey("correlationKey")
+        .correlationKey("")
         .variables(variables)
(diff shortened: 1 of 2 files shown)


saig0 (approved):
@pihme looks good :+1:

pihme:
bors merge

PR #656, opened 2022-04-22
feat: add zeebe version to failure notification

### LINKED ISSUE #411: Improve testbench fail message such that it shows the tested version
Most of the time when testbench fails and sends a message to slack I first need to find out which version was affected, it would my life easier if the error message contains it. Sometimes the link to the ci build contains the version, but it is hard to detect.

**Current message:**

```
:bpmn-error-throw-event:
chaosToolkit on Production - M failed for generation temp_kVEcr9UVh1 on cluster signal-inclusive-task
<null|Operate>
Business Key:https://ci.zeebe.camunda.cloud/job/camunda-cloud/job/zeebe/job/stable%252F0.26/150/
There were 1 failures.
Action with name: 'Deploy different deployment versions.' timed out after PT1H.
Chaos experiment 'Zeebe deployment distribution' failed.
Experiment failed for cluster plan: 'Production - M' against cluster 'signal-inclusive-task' (ccd419a8-1aa9-4e56-8e4c-418938fcac29).
```

deepthidevaki:
closes #411

### DIFF
--- a/core/src/main/java/io/zeebe/clustertestbench/handler/NotifyEngineersHandler.java
+++ b/core/src/main/java/io/zeebe/clustertestbench/handler/NotifyEngineersHandler.java
@@ -37,17 +37,18 @@ public class NotifyEngineersHandler implements JobHandler {
     resultBuilder.append("\n");
 
     // message
-    resultBuilder
-        .append("_")
-        .append(input.getTestProcessId())
-        .append("_") //
-        .append(" on ")
-        .append("_" + input.getClusterPlan() + "_") //
-        .append(" failed for generation ")
-        .append("`" + input.getGeneration() + "`") //
-        .append(" on cluster ")
-        .append("_" + input.getClusterName() + "_")
-        .append(" `" + input.getClusterId() + "`");
+    final var errorMessage =
+        String.format(
+            "_%s_ on _%s_ failed for branch `%s`. %n Zeebe image: _%s_ %n Generation: _%s_ %n Cluster : _%s_ `%s`",
+            input.getTestProcessId(),
+            input.getClusterPlan(),
+            input.getBranch(),
+            input.getZeebeImage(),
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/zeebe/clustertestbench/handler/NotifyEngineersHandler.java
```diff
@@ -37,17 +37,32 @@ protected String composeMessage(final Input input) {
...
+        .append(input.getClusterPlan())
+        .append("_") //
         .append(" failed for generation ")
-        .append("`" + input.getGeneration() + "`") //
+        .append("`")
+        .append(input.getGeneration())
+        .append("`") //
+        .append(" using Zeebe version ")
+        .append("`")
+        .append(input.getZeebeImage())
```
korthout, line 57: Note that there are different properties that also provide version information, like the `generation` (although this will be temp_xxx for testbench clusters) and `branch`.

🔧 I'd prefer to not call it Zeebe version, but rather Zeebe image. It might be confusing otherwise. 

🔧 We might benefit from also adding the branch information to the message

korthout (approved):
Nice addition @deepthidevaki Thanks!

I have a few suggestions, but LGTM 👍

core/src/main/java/io/zeebe/clustertestbench/handler/NotifyEngineersHandler.java
deepthidevaki: I will add the branch too.

deepthidevaki:
Thanks for you review @korthout. I changed the error message because now we have a lot more info. If you want please take a look at it again, otherwise I will merge it later today.

korthout:
Looks good @deepthidevaki

🔧 You could still add the branch to the test case.

deepthidevaki:
bors merge

PR #754, opened 2023-02-01
Add workflow used by github actions to trigger qa

deepthidevaki:
This replaces the use of external-tool-integration. It uses slack connector to notify success/failure instead of explicit fetch result task. This simplifies the github workflow a lot. (See https://github.com/camunda/zeebe/pull/11508)

The process is already deployed and tested. You can view it [here](https://bru-3.operate.ultrawombat.com/7b2cd3f5-cfcc-4a65-ab22-0ccaa071e68e/processes?process=qa-github-trigger&version=all&active=true&incidents=true&completed=true&canceled=true)


Zelldon:
Thanks @deepthidevaki can we then remove the external trigger process? :thinking:

deepthidevaki:
> can we then remove the external trigger process?

May be. I was not sure if it is used anywhere. Or if it will be useful.

Zelldon (approved):
Thanks :)

PR #770, opened 2023-03-02
fix: don't filter cluster plan by region

### LINKED ISSUE #768: Can't lookup `k8sContext` by cluster plan type anymore
Due to changes here: https://github.com/camunda-cloud/console/issues/3935, looking up the k8sContext by cluster plan type does not work anymore. The `k8sContext` is now available as `regions` and is not coupled to a cluster plan type.

This breaks at least this access: https://github.com/zeebe-io/zeebe-cluster-testbench/blob/aac020c0ca033a4dcb7c3f966e51d3e72ca8404b/core/src/main/java/io/zeebe/clustertestbench/handler/MapNamesToUUIDsHandler.java#L131-L134

```
java.lang.NullPointerException: Cannot invoke "io.zeebe.clustertestbench.cloud.CloudAPIClient$K8sContextInfo.uuid()" because the return value of "io.zeebe.clustertestbench.cloud.CloudAPIClient$ParametersClusterPlanTypeInfo.k8sContext()" is null
```

OpenAPI spec: https://api.cloud.ultrawombat.com/openapi/docs/#/default/GetParameters

oleschoenburg:
This association does not exist anymore. Cluster plan types are independent from the region so we don't have to filter anymore. This fixes a bug where the cluster plan type does not provide the k8sContext anymore.

fixes #768


megglos (approved):
LGTM!

PR #772, opened 2023-03-02
fix(ci): add git safe directory for release

megglos:
This addresses a new safety check that currently leads to a failing release.

See similar fix to other Jenkins setup [here](https://github.com/camunda/camunda-optimize/commit/8e76b89f3272a1dd5144ed200a84c63ba54c1a19).


oleschoenburg (approved):
Nice, thanks!

PR #778, opened 2023-03-10
fix: don't hide incident when QA times out

oleschoenburg:
Use a non-interrupting boundary timer event instead of an interrupting one to ensure that incidents in the called process remain visible in Operate.


Zelldon (approved):
Thanks @oleschoenburg :bow:

PR #789, opened 2023-03-30
fix: move process to correct location

deepthidevaki:
The process was incorrectly added to a different location. So changes to it are not automatically deployed to testbench.


korthout (approved):
LGTM

korthout:
bors merge

PR #801, opened 2023-04-13
Change region from "new chaos" to "Belgium, Europe (europe-west1) aka. new chaos"

remcowesterhoud:
Console changed the name of this region.
Because of this the process runs into incidents, as it cannot find the expected "new chaos" region.


koevskinikola (approved):
LGTM!

bors merge

Issue #583, opened 2022-03-07
Replace Jenkins with GitHub Actions


pihme:

pihme:
duplicate of #649

PR #826, opened 2023-05-09
fix(processes): rename used region for QA runs

korthout:
The region name has changed:
- from `Belgium, Europe (europe-west1) aka. new chaos`
- to `Belgium, Europe (europe-west1)`


remcowesterhoud (approved):
Thanks @korthout 

💭 We should probably move to region UUID instead of names 😄

korthout:
Seems bors is broken since:
- #583 

I'll open a pull request to fix it

korthout:
bors merge

PR #827, opened 2023-05-09
Switch bors config to await github workflow

korthout:
Recently Jenkins has been swapped out for GHA, but it seems Bors' configuration was not considered.
- #583 

This commit applies the same structure used in the Zeebe repo to allow Bors to await a GHA workflow that awaits one or several GHA jobs.


remcowesterhoud (approved):
I sure do

bors merge

PR #830, opened 2023-05-10
Add slack webhook url as environment variable

remcowesterhoud:
- The `ZCTB_SLACK_WEBHOOK_URL` env var is an empty string
- The var is required, but empty string is accepted
- This env var is filled from a k8s secret, other secrets are correctly filled in
- The k8s secret is also an empty string
- The deploy.sh script is run from build.yml workflow but does not pass in the url as secret (even though the secret is retrieved in the import secrets step)


korthout (approved):
Thanks @remcowesterhoud 💐 

LGTM 👍 

bors merge

PR #860, opened 2023-07-14
refactor(processes): increase retry delay when querying cluster state

deepthidevaki:
The timer to retry "Query Zeebe Cluster State in Camunda Cloud" is increased from 10s to 1 minute.
image

Sometimes we get rate limited by cloud api resulting in incidents and test failures. To reduce the likelihood of hitting 'Too many requests' error we reduce the retry frequency.


npepinpe (approved):
:+1:

deepthidevaki:
bors merge

Issue #889, opened 2023-09-14
Race condition on creation of clusters 


Zelldon:
We have now reappearing cluster creation failures, due to a race condition when creating the generation and updating the related channel.

The symptom we observe is that there is an incident in testbench that a cluster can't be created. 

A potential reason is that the channel update fails, because the generation is not yet created. But this is not transparent from outside.

[In testbench logs we see something like this, when cluster creation fails](https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22zeebe-io%22%0Aresource.labels.location%3D%22europe-west1-b%22%0Aresource.labels.cluster_name%3D%22zeebe-cluster%22%0Aresource.labels.namespace_name%3D%22testbench-1-x-prod%22%0Alabels.k8s-pod%2Fapp%3D%22testbench%22;pinnedLogId=2023-09-14T09:41:46.053105684Z%2Ff3opoemayigfe95i;cursorTimestamp=2023-09-14T09:41:46.053105684Z;duration=PT15M?project=zeebe-io)

```
Expected to create cluster compensation-exclusive-transaction, but failed

jakarta.ws.rs.client.ResponseProcessingException: java.io.IOException: BAD REQUEST returned from URI: https://api.cloud.ultrawombat.com/clusters/, requestBody:{"name":"compensation-exclusive-transaction","planTypeId":"0282d9dd-266f-4db6-a451-724ff62648f4","channelId":"cf355219-cc2d-4266-83e1-2430998ddf30","generationId":"9728580a-744c-4bb4-a771-e4dc3b859c49","regionId":"cab01f7b-0b95-42a3-83d0-b66d53fd5255"}, responseBody: Generation is not part of allowed generations of selected channel.
	at org.jboss.resteasy.client.jaxrs.internal.ClientInvocation.filterResponse(ClientInvocation.java:671) ~[testbench.jar:3.0.5-SNAPSHOT]
	at org.jboss.resteasy.client.jaxrs.internal.ClientInvocation.invoke(ClientInvocation.java:428) ~[testbench.jar:3.0.5-SNAPSHOT]
	at org.jboss.resteasy.client.jaxrs.internal.proxy.ClientInvoker.invokeSync(ClientInvoker.java:134) ~[testbench.jar:3.0.5-SNAPSHOT]
[...]

PR #891, opened 2023-09-18
Log more response errors

Zelldon:
Previously we only logged (not sure but current assumption) BAD_REQUEST returned from SAAS

Related to https://github.com/zeebe-io/zeebe-cluster-testbench/issues/889


cloud-client/src/main/java/io/zeebe/clustertestbench/cloud/filter/BadResponseFilter.java
```diff
@@ -7,26 +7,34 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(BadResponseFilter.class);
 
   @Override
   public void filter(
       final ClientRequestContext requestContext, final ClientResponseContext responseContext)
       throws IOException {
+    if (responseContext.getStatus() > 399) {
 
-    if (responseContext.getStatus() == 400) {
       final String requestBody = new ObjectMapper().writeValueAsString(requestContext.getEntity());
```
megglos, line 23: 🔧  pre-existing and minor given the context but while we are here anyway, could just make the objectmapper a static field to initialize once and reuse as it's relatively expensive to construct

core/src/test/java/io/zeebe/clustertestbench/handler/FailedServiceShouldFailResponseTest.java
```diff
@@ -0,0 +1,62 @@
...
+    client = ClientBuilder.newBuilder().register(BadResponseFilter.class).build();
+    final var target = (ResteasyWebTarget) client.target("http://localhost:8081");
+    final var cloudAPIClient = target.proxy(InternalCloudAPIClient.class);
+
+    spyInternalApiClient = spy(cloudAPIClient);
+    localhost = HttpServer.create(new InetSocketAddress("localhost", 8081), 1);
+    localhost.start();
+    activatedJobStub = jobClientStub.createActivatedJob();
+    sutCreateGenerationHandler = new CreateGenerationInCamundaCloudHandler(spyInternalApiClient);
+  }
```
megglos, line 46: 💭 I wonder if we could make the test setup simple, maybe like resteasy does response filter testing here https://github.com/resteasy/resteasy/blob/main/testsuite/unit-tests/src/test/java/org/jboss/resteasy/test/client/ClientResponseFilterTest.java but also fine to keep it that way

megglos (approved):
LGTM! Just some ideas/suggestions.

core/src/test/java/io/zeebe/clustertestbench/handler/FailedServiceShouldFailResponseTest.java
Zelldon: Actually, I wrote the test, due to lack of knowledge and I wanted to verify how the client behaves when there is a service but doesn't answer or returns an error. This was not clear nor transparent to me. But I guess we can also remove it again.

PR #975, opened 2024-01-11
ci: migrate to workload federation

Zelldon:
Recently I removed old services accounts from our IAM, see https://app.slack.com/client/T0PM0P1SA/C064CNSFKFF

This caused to fail the builts/deployments for testbench.

This PR migrates the old Jenkins service account JSON usage to the official workload identity federation usage, [which is recommended by Google](https://github.com/google-github-actions/auth?tab=readme-ov-file#workload-identity-federation-through-a-service-account).

I have set up a new service account and created a new workload identity pool, etc. Since I know had to do this multiple times already I wrote a script, which we can reuse the next time

```sh
#!/bin/bash
gcloud components update

# https://github.com/google-github-actions/auth?tab=readme-ov-file#workload-identity-federation-through-a-service-account

name="testbench"
detailedName="ZBTestbench"
org=zeebe-io
repo=zeebe-io/zeebe-cluster-testbench

export PROJECT_ID="zeebe-io"

gcloud iam service-accounts create "$name-service-account" \
  --project "${PROJECT_ID}"

gcloud iam workload-identity-pools create "$name-identity-pool" \
  --project="${PROJECT_ID}" \
  --location="global" \
  --display-name="$detailedName identity pool"

export WORKLOAD_IDENTITY_POOL_ID=$(gcloud iam workload-identity-pools describe "$name-identity-pool" \
  --project="${PROJECT_ID}" \
  --location="global" \
  --format="value(name)")

gcloud iam workload-identity-pools providers create-oidc "$name-identity-provider" \
  --project="${PROJECT_ID}" \
  --location="global" \
  --workload-identity-pool="$name-identity-pool" \
  --display-name="$detailedName identity provider" \
  --attribute-mapping="google.subject=assertion.sub,attribute.actor=assertion.actor,attribute.repository=assertion.repository,attribute.repository_owner=assertion.repository_owner,attribute.workflow=assertion.workflow,attribute.ref=assertion.ref" \
  --issuer-uri="https://token.actions.githubusercontent.com" \
  --attribute-condition="attribute.repository_owner=='$org'"

gcloud iam workload-identity-pools providers describe "$name-identity-provider" \
  --project="${PROJECT_ID}" \
  --location="global" \
  --workload-identity-pool="$name-identity-pool" \
  --format="value(name)"

gcloud iam service-accounts add-iam-policy-binding "$name-service-account@${PROJECT_ID}.iam.gserviceaccount.com" \
  --project="${PROJECT_ID}" \
  --role="roles/iam.workloadIdentityUser" \
  --member="principalSet://iam.googleapis.com/${WORKLOAD_IDENTITY_POOL_ID}/attribute.repository/${REPO}"

```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,10 +52,8 @@ jobs:
         secretId: ${{ secrets.VAULT_SECRET_ID }}
         exportEnv: false # we rely on step outputs, no need for environment variables
         secrets: |
-          secret/data/products/zeebe/ci/jenkins ZEEBE_JENKINS_DEPLOY_SERVICEACCOUNT_JSON;
           secret/data/products/zeebe/ci/zeebe ARTIFACTS_USR;
           secret/data/products/zeebe/ci/zeebe ARTIFACTS_PSW;
-          secret/data/products/zeebe/ci/zeebe ZEEBE_GCR_SERVICEACCOUNT_JSON;
           ${{ env.vault_secret_path }} clientSecret;
           ${{ env.vault_secret_path }} cloudClientSecret;
           ${{ env.vault_secret_path }} contactPoint;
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -105,34 +103,39 @@ jobs:
       env:
         MAVEN_USERNAME: ${{ steps.secrets.outputs.ARTIFACTS_USR }}
         MAVEN_PASSWORD: ${{ steps.secrets.outputs.ARTIFACTS_PSW }}
-
-    - name: Login to Docker Registry
+    - uses: google-github-actions/auth@v2
+      if: env.should_deploy_infra == 'true'
```
Zelldon, line 107: I dislike a bit that we have to use all the ifs, in the workflow, but I didn't want to spent more time to refactor it and focus on one change/fix for now.

npepinpe (approved):
:tada:

.github/workflows/build.yml
npepinpe: Yeah it's fine, we don't touch it that often anyway

npepinpe: Is this the last place we use these secrets? Should we tell infra to remove them? :thinking:

nicpuppa (approved):
💪

.github/workflows/build.yml
Zelldon: I think so, the secrets where actually coming from our IAM so here I deleted it already. But I guess you referring to vault right? I guess I can delete it as well there.

PR #980, opened 2024-01-15
ci: dont specify specific maven version

korthout:
It's unclear to me why these jobs specify a specific maven version. It's not managed by any version manager, so it quickly gets out of date.

It was introduced with https://github.com/zeebe-io/zeebe-cluster-testbench/pull/802

Instead, it's fine to use the latest maven version for our CI. This is needed because the flatten maven plugin v1.6 requires a newer maven version then the one we currently specify in our CI.

This will help us merge:
- https://github.com/zeebe-io/zeebe-cluster-testbench/pull/977

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,7 +71,6 @@ jobs:
       with:
         java-version: '21'
         distribution: 'temurin'
-        maven-version: '3.6.2'
 
     # Use CI Nexus as co-located pull-through cache for Maven artifacts via ~/.m2/settings.xml
     - name: Create Maven Settings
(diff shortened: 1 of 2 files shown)


Zelldon (approved):
> It's unclear to me why these jobs specify a specific maven version. It's not managed by any version manager, so it quickly gets out of date.

This is better practice to have reproducible builds, of course, this only works if we have also renovate or something up set that updates such dependencies as well.


Approving unblocking you here, but we likely should also migrate to renovate here.

korthout:
Thanks for the review @Zelldon 🙇 

>we likely should also migrate to renovate here.

Renovate is enabled and configured here, but I'm not sure it can update a version specific as action input.

Anyways, we also don't use this in the Zeebe repo AFAIK: https://github.com/camunda/zeebe/blob/main/.github/actions/setup-zeebe/action.yml

So, let's go: bors merge

korthout:
bors merge

PR #982, opened 2024-01-17
refactor(core): remove guard clause for generation cloning

### LINKED ISSUE #981: Remove guard clause for generation cloning
### Description

A guard clause was introduced in the testbench code to prevent the https://github.com/camunda-cloud/camunda-cloud-management-apps/issues/1731 bug from occurring.

### AC
As the issue has been fixed, the code needs to be removed.

koevskinikola:
A bug in Camunda Cloud Console prevented cloning of generations that were not marked as upgradable from. The issue has been fixed and the guard clause needs to be removed from the testbench code.

closes #981

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -43,6 +43,9 @@
     <MarkdownNavigatorCodeStyleSettings>
       <option name="RIGHT_MARGIN" value="72" />
     </MarkdownNavigatorCodeStyleSettings>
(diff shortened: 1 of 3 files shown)


.idea/codeStyles/Project.xml
koevskinikola, line 1: :thought_balloon: Seems like this code was generated by IntelliJ. I'll revert it asap.

Zelldon (approved):
Great thank you!

PR #1030, opened 2024-04-09
Improve failure messages

Zelldon:
Adds some better formatting, improved wording (speaking of target cluster) on the failure messages and adds the Operate URL to easily navigate to the orchestration cluster.

I will keep improving this, like adding logs URL, etc. but will do this later.

Example message:

```
:bpmn-error-throw-event:
_processID_ on _CLUSTER_PLAN_B_ failed for branch .
Check out here: http://url/clusterId

*Details:*

* Zeebe image: _ZEEBE:VERSION_
* Generation: _GENERATION_C_
* Target cluster : _CLUSTER_A_
* Target cluster Operate: https://localhost.test/
* Business key: http://jenkins/branch/build

*Failures:*

Failure count: 1

Error message 1
Error message 2
```

### DIFF
--- a/core/src/main/java/io/zeebe/clustertestbench/bootstrap/Launcher.java
+++ b/core/src/main/java/io/zeebe/clustertestbench/bootstrap/Launcher.java
@@ -282,6 +282,25 @@ public class Launcher {
         Duration.ofSeconds(10));
   }
 
+  protected static String convertZeebeUrlToOperateUrl(final String endpoint) {
+    // Zeebe GRPC endpoint looks normally like this:
+    // <clusterId>.bru-2.zeebe.camunda.io/
+    //
+    // Operate looks like this:
+    // bru-2.operate.camunda.io/<clusterId>
+    //
+
+    // removing potential protocol from endpoint
+    var strippedEndpoint = endpoint.replace("http://", "");
+    strippedEndpoint = strippedEndpoint.replace("https://", "");
+
+    // finding index when clusterId stops
+    final int firstDotIndex = strippedEndpoint.indexOf('.');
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/zeebe/clustertestbench/bootstrap/Launcher.java
```diff
@@ -239,10 +239,15 @@ private void registerWorkers(final ZeebeClient client) {
         client, "run-sequential-test-job", new SequentialTestHandler(), Duration.ofMinutes(30));
 
     final var slackNotificationService = new SlackNotificationService(slackWebhookUrl);
+
+    // Testbench cluster Operate url
+    final String operateUrl =
+        convertZeebeUrlToOperateUrl(testOrchestrationAuthenticatonDetails.getServerURL());
```
oleschoenburg, line 245: :thought_balloon: I was not able to follow the code and verify that `testOrchestrationAuthenticatonDetails` is the right one to use here. I'm assuming you either tested this or are otherwise sure that this is the right one?

oleschoenburg (approved):
Nice improvement, thanks! :medal_sports:

core/src/main/java/io/zeebe/clustertestbench/bootstrap/Launcher.java
Zelldon: Lol good that you pointed this out it was actually the audience not the serverUrl 👍🏼 https://github.com/zeebe-io/zeebe-cluster-testbench/blob/main/core/src/main/java/io/zeebe/clustertestbench/bootstrap/BootstrapFromEnvVars.java#L34

PR #1031, opened 2024-04-09
feat: add zbchaos console log url

Zelldon:
When failing chaos experiments we have a long way to find the right logs, especially if we want to look at the zbchaos logs, as this is not linked.

With this PR we close the gap and have a direct connection to the running pods logs


oleschoenburg (approved):
:star_struck:

PR #1032, opened 2024-04-09
Reduce indirection for QA trigger

Zelldon:
This PR reduces the complexity of our process models (by factor 5) and need of tests clusters by 50%.

We have one abstraction/call-activity layer:

qa-github-trigger

**Before:**

The QA trigger process model was quite abstract, and going to find the right chaos orchestration was several layers hidden behind call activities.

We needed 5+ process models to reach the actual orchestraction of chaos experiments.

QA-Github-trigger
step1
QA-protocol
step2
run-all-tests-in-camunda-cloud-per-cluster-plan
step3
run-all-tests-in-camunda-cloud
step4
run-test-in-camunda-cloud
step5

**Details:**

This PR should reduce the indirection and clean up some unnecessary abstractions.

- Removed/inlined several call activities that had no real use.
- For example, the process run all tests (the name itself was confusing as it shared the prefix with another one), and we never ran against several cluster plans. (only at the begin).
- The change should make it easier to understand what is going on and where to find the correct execution.
- with the refactoring we now instead of creating two clusters, per run we now have one.
  - First sequential tests are executed (which are not much) and afterward the chaos experiments, that should be fine and reduces cost by 50 %. Especially it reduced another abstraction.


Zelldon:
@oleschoenburg  

I can't test it currently in depth via dev set up or anything, so I would propose a more async method.

 Meaning we merge it, and I trigger some runs to see whether it is working as expected (it is anyway broken right now) wdyt? I will fix it if I see some issues.

oleschoenburg:
Yeah that sounds good to me!

PR #1033, opened 2024-04-11
Clean up Testbench repo

Zelldon:
# Description

Remove several unused process models and documentation after they have been merged together, see https://github.com/zeebe-io/zeebe-cluster-testbench/pull/1032.

## Details

We removed the onion principle from testbench.

In general this principle makes sense, but as with everything
until a certain limit and for certain use cases.
Like if you have the use case to reuse components, want better
separation of concerns, improve flexibility, etc.

In our case with testbench it was unfortunatelty not really useful and
needed, as we never had the use case of running only separate processes,
especially because you had to provide the necessary variables as
well (which is not obvious or hard to provide).

Furthermore, it made it too deep and complex. Due to the complexity
it was not easy to understand, track or
debug.

It shouldn't be necessary to have a separate documentation to understand
the models and how the data flows.


oleschoenburg (approved):
:heart_eyes:

PR #1121, opened 2024-10-22
feat: use REST for QA

nicpuppa:


megglos (approved):
Let's try it and inform the medic about the change as QA might be affected tomorrow ^^

